PR #110, opened 2020-04-01
Single run revamp

tohardik:
Major changes

1. Removed String concatenation based result storage, replaced with Jena graph
2. Changes in benchmark result: fixed typos, removed term collisions, introduced new edges
3. Replaced positional arguments based worker configuration to key value pair based arguments
4. Revamped result processing for SPARQLWorker(Processing in background threads, caching of already processed results)
5. Added feature to override response type of SPARQL query from the configuration
6. Improved precision of time measurement from microseconds to nanoseconds
7. New class hierarchy for Worker classes
8. Several bug fixes


bigerl:
@TortugaAttack @tohardik Have we anything left that should be addressed before merging and creating version 3.0.0?

TortugaAttack:
Is there an overview on whats different or new  (resp. a changelog)?

Otherwise i would say go on merge :)

tohardik:
@bigerl On second thought, I'll merge it after I push out a few more changes which address the different ways to add/process the results between Http and Cli based workers(we discussed this last time on call).

iguana.corecontroller/src/main/java/org/aksw/iguana/cc/config/IguanaConfig.java
```diff
@@ -145,37 +152,45 @@ private static void addRecursive(Configuration target, Configuration source, Str
...
-			return null;
-		}
-		return id;
+		// Suite ID collision fix start
+		int currentTimeMillisHashCode = Math.abs(Long.valueOf(Instant.now().getEpochSecond()).hashCode());
+		return String.valueOf(currentTimeMillisHashCode);
+		// Suite ID collision fix end
+
+
+//		File suiteIDFile = new File("suite.id");
```
bigerl, line 171: why not remove the code?

PR #111, opened 2020-04-09
Use HTTPS to resolve dependencies in Maven Build

LukasBluebaum:
Resolve dependencies of the AKSW repository over https.


bigerl:
Makes sense. Any concerns? @TortugaAttack @tohardik

PR #128, opened 2020-12-08
Feature/support HTTP chunked encoding (Streaming) 

bigerl:


bigerl:
@TortugaAttack 
we might need to reconsider this: https://www.baeldung.com/httpclient-timeout

Otherwise this could block for longer than timeOut duration here: https://github.com/dice-group/IGUANA/blob/13c6c6c60221c88e482330d615a769eed97aca9e/iguana.corecontroller/src/main/java/org/aksw/iguana/cc/worker/impl/HttpWorker.java#L95

TortugaAttack (approved):

PR #130, opened 2020-12-10
changing streaming structure, adding throwawaylangproc

TortugaAttack:


iguana.commons/src/main/java/org/aksw/iguana/commons/streams/Streams.java
```diff
@@ -40,4 +40,30 @@ static public String inputStream2String(InputStream inputStream, Instant startTi
...
     }
+
+    static public int inputStream2String(InputStream inputStream, Instant startTime, double timeout, StringBuilder response) throws IOException, TimeoutException {
+        ByteArrayOutputStream result = new ByteArrayOutputStream();
+        byte[] buffer = new byte[10 * 1024 * 1024]; // 10 MB buffer
+        int length;
+        while ((length = inputStream.read(buffer)) != -1) {
+            if (durationInMilliseconds(startTime, Instant.now()) > timeout)
+                throw new TimeoutException("reading the answer timed out");
+            result.write(buffer, 0, length);
```
TortugaAttack, line 51: maybe add here to StringBuilder directly, or use BAOS instead of StringBuilder

TortugaAttack: changed to BAOS

iguana.commons/src/main/java/org/aksw/iguana/commons/streams/Streams.java
```diff
@@ -14,30 +14,100 @@
...
+
+    /**
+     * Fastest way to serialize a stream to UTF-8 according to https://stackoverflow.com/a/35446009/6800941
+     *
+     * @param inputStream the stream to read from
+     * @param result      the stream where the result is written to.
+     * @return size of the output stream
+     * @throws IOException from inputStream.read
+     */
+    public static int inputStream2ByteArrayOutputStream(InputStream inputStream, ByteArrayOutputStream result) throws IOException {
```
TortugaAttack, line 82: use 
```java
int inputStream2ByteArrayOutputStream(InputStream inputStream, ByteArrayOutputStream result)throws IOException {
return inputStream2ByteArrayOutputStream(inputStream, Instant.now(), -1, result);
}
```

iguana.corecontroller/src/main/java/org/aksw/iguana/cc/lang/impl/SPARQLLanguageProcessor.java
```diff
@@ -124,9 +123,9 @@ public static String getContentTypeVal(Header header) {
 
     public static long getJsonResultSize(String res) throws ParseException {
```
TortugaAttack, line 124: This will become terrible to maintain. 
Also just to get the result size. Maybe we should remove the result size feature. 
This is too much hassle for too less output.

TortugaAttack: Esp as this is only for SPARQL, not even generic

iguana.commons/src/main/java/org/aksw/iguana/commons/streams/Streams.java
```diff
@@ -14,30 +14,100 @@
...
      *
-     * @param inputStream stream to be serialized
-     * @return content of the inputStream as string
-     * @throws IOException forwarded exception from InputStream.read()
+     * @param inputStream the stream to read from
+     * @return the content of inputStream as a string.
+     * @throws IOException from inputStream.read
      */
     static public String inputStream2String(InputStream inputStream) throws IOException {
         ByteArrayOutputStream result = new ByteArrayOutputStream();
```
TortugaAttack, line 24: We need to switch to BigBAOS (will work on that today&tomorrow)

TortugaAttack (commented):
Yeah, we need to change a bit and discuss if we want to remove the resultSize into the byte result size instead, or remove it in general.

iguana.commons/src/main/java/org/aksw/iguana/commons/streams/Streams.java
TortugaAttack: too many complications for this special case

iguana.commons/src/main/java/org/aksw/iguana/commons/io/BigByteArrayInputStream.java
```diff
@@ -0,0 +1,54 @@
...
+        curArray=bbaos.getBaos().get(curPos++).toByteArray();
+    }
+
+    @Override
+    public int read() throws IOException {
+        if(eos()){
+            return -1;
+        }
+        int ret;
+        //Is it Integer.MAX_VALUE or Integer.MAX_VALUE-1?
```
TortugaAttack, line 35: @bigerl 
I cannot wrap my head around it atm. Can you confirm it?

bigerl: must be `Integer.MAX_VALUE - 3`

iguana.commons/src/main/java/org/aksw/iguana/commons/io/BigByteArrayOutputStream.java
```diff
@@ -0,0 +1,108 @@
...
+
+    @Override
+    public void write(int i) throws IOException {
+        ByteArrayOutputStream current = baos.get(baos.size() - 1);
+        current = ensureSpace(current);
+        current.write(i);
+    }
+
+    private ByteArrayOutputStream ensureSpace(ByteArrayOutputStream current) {
+        if (current.size() == 2147483639) {
```
bigerl, line 78: What's that number?
I would have expected
2147483645 = Integer.MAX_VALUE - 2 not
2147483639

PR #180, opened 2022-09-23
fix accepting all 200-299 status codes on http responses

frensing:
All status codes from 200 to 299 will be accepted as successful.
Is a test needed? I am not sure on how to best implement this.


bigerl (approved):

bigerl:
The fix was good and working. I just wanted to take the chance and simplify the code a bit. It would have taken much longer to write what I wanted to point out then adjusting it myself. 
- [ ] Please double check my changed that they did not change the behavior of your fix.

Issue #191, opened 2023-03-22
FileUtils method readLineAt doesn't properly read windows-style line endings


nck-mlcnv:
When the file, that is being read, uses `\r\n` as its line endings, the method doesn't remove the carriage return. Right now, it mainly causes issues with the tests.

PR #197, opened 2023-04-08
Fix for #191

nck-mlcnv:
This pull request also does the following:

- Fix the same issue with the `countLines` method, although it's a quite simple fix for that method.
  - For this I've added the `getLineEnding` method, which returns the used line separator of a file.
- Add better tests.


bigerl:
please resolve conflicts

iguana.corecontroller/src/main/java/org/aksw/iguana/cc/utils/FileUtils.java
```diff
@@ -100,6 +107,64 @@ public static String readFile(String path) throws IOException {
...
+	 * @param filepath this string that contains the path of the file
+	 * @return the line ending used in the given file
+	 * @throws IOException
+	 */
+	public static String getLineEnding(String filepath) throws IOException {
+		int lineEndingIndex = 0;
+		try(BufferedReader br = new BufferedReader(new FileReader(filepath))) {
+			// readline consumes the line endings mentioned in the javadoc, thus the length of a line equals the index
+			// of the line's ending
+			lineEndingIndex = br.readLine().length();
```
bigerl, line 130:
```
char b;
while ((b = (char) br.read()) != -1) {
			// check if b is // \r or \n, if \r check if \n is the next. 
		}
```

should do it

iguana.corecontroller/src/main/java/org/aksw/iguana/cc/utils/FileUtils.java
```diff
@@ -100,6 +107,64 @@ public static String readFile(String path) throws IOException {
...
+		Pattern pattern = Pattern.compile("\\R");
+		Matcher matcher = pattern.matcher(bufferString);
+		if(matcher.find()) {
+			return matcher.group();
+		}
+		else {
+			// if for some reason, the matcher still doesn't find a line ending
+			return System.lineSeparator();
+		}
+
```
bigerl, line 167: here is a `}` missing.

iguana.corecontroller/src/test/java/org/aksw/iguana/cc/utils/FileUtilsTest.java
```diff
@@ -29,6 +51,22 @@ public void readLineAtTest() throws IOException {
         assertEquals("dfe", FileUtils.readLineAt(4, f));
         //read line at -1
         assertEquals("", FileUtils.readLineAt(-1, f));
+
+        File f1 = new File("src/test/resources/readLineTestFile1.txt");
+        File f2 = new File("src/test/resources/readLineTestFile2.txt");
```
bigerl, line 56: please separate the processing of the files. Additionally, define the mappings sth like this: 
```
Map<String, Integer> left = ImmutableMap.of(
0, "line 1",
1, "line 2", 
... )
```

I would suggest to define a lambda function that get's such a map and a file path as input and preforms all the checks. Then you can call the lambda for the 4 files tested in this test case. 

Alternatively it would also make sense to break them down into sub-test cases. Please investigate what the possiblities are here. I am not into junit. 

In the report it should roughly look like this:
```
✅ read line at
  ✅ fileUtils.txt
  ✅ readLineTestFile1.txt
  ✅ readLineTestFile2.txt
  ✅ readLineTestFile3.txt
```

iguana.corecontroller/src/test/java/org/aksw/iguana/cc/utils/FileUtilsTest.java
```diff
@@ -29,6 +51,22 @@ public void readLineAtTest() throws IOException {
         assertEquals("dfe", FileUtils.readLineAt(4, f));
         //read line at -1
         assertEquals("", FileUtils.readLineAt(-1, f));
```
bigerl, line 53: I think this should throw. Or is there any valid use-case for this?

bigerl (changes requested):
Thanks for the PR. I think the code that was used here before is way too over complicated and mostly beyond fixing. It needs more or less a complete rewrite of the respective methods. I suggested some directions in my comments.

iguana.corecontroller/src/main/java/org/aksw/iguana/cc/utils/IndexedLineReader.java
```diff
@@ -0,0 +1,189 @@
...
+     *  files that are roughly 2GB big.)*/
+    private ArrayList<Integer> indices;
+
+    private String filepath;
+
+    /** Stores the filesize in number of bytes. */
+    private int filesize;
+
+    /** If no line separator is given through the constructor, this string stays empty. */
+    private String separator = "";
```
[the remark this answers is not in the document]
bigerl: Die richtig saubere Variante wäre: https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html

iguana.corecontroller/src/test/java/org/aksw/iguana/cc/utils/FileUtilsTest.java
nck-mlcnv: IndexedLineReader now throws an exception, when a line doesn't exist with a specified index.

iguana.corecontroller/src/main/java/org/aksw/iguana/cc/utils/FileUtils.java
```diff
@@ -14,74 +16,79 @@
...
-	 * @return line at pos
+	 * Returns a line at a given position of a File. <br/>
+	 * This method ignores every empty line, therefore the parameter <code>pos</code> references the n-th non-empty line.
+	 *
+	 * @param index the position of a non-empty line which should be returned
+	 * @param file 	the file to read from
+	 * @return the line at the given position
 	 * @throws IOException
 	 */
-	public static String readLineAt(int pos, File filename) throws IOException {
```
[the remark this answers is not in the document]
nck-mlcnv: See IndexedLineReader

iguana.corecontroller/src/main/java/org/aksw/iguana/cc/utils/IndexedLineReader.java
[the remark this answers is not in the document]
nck-mlcnv: changed in new commit

iguana.corecontroller/src/main/java/org/aksw/iguana/cc/utils/IndexedLineReader.java
```diff
@@ -93,52 +117,103 @@ public int size() {
...
+
+            if(raf.getFilePointer() >= nextLinePosition ||
+                    (this.separator.get().isBlank() && currentLine.isBlank())) {
+                // Break the while-loop, when the line, that contains the separator has been read, or if the current
+                // line is blank, when the separator is blank.
+                break;
+            }
+        }
+        output.append(lastLine);
+        return output.toString();
```
bigerl: I would suggest something like this: 
```
RandomAccessFile raf = new RandomAccessFile(this.filepath, "r");
byte[] buffer = new byte[this.indices.get(index)[1] - this.indices.get(index)[0]];
raf.read(buffer, this.indices.get(index)[0], this.indices.get(index)[1] - this.indices.get(index)[0]);
return new String(buffer, StandardCharsets.UTF_8);
```

iguana.corecontroller/src/main/java/org/aksw/iguana/cc/utils/FileUtils.java
```diff
@@ -125,9 +124,13 @@ public static String readFile(String path) throws IOException {
 	public static String getLineEnding(String filepath) throws IOException {
```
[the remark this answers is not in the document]
bigerl: actually, the file reader should be managed by a try block here. please patch when integrating the code.

iguana.corecontroller/src/main/java/org/aksw/iguana/cc/utils/IndexedLineReader.java
bigerl: This access pattern would actually make sense to be used everywhere

iguana.corecontroller/src/main/java/org/aksw/iguana/cc/utils/IndexedLineReader.java
```diff
@@ -5,67 +5,91 @@
...
+ *     <li>a custom separator</li>
  * </ul>
  * This class does not index blank lines.
  */
 public class IndexedLineReader {
 
     /** This list stores the indices of the bytes, at which a new line starts. (Stores Integers, so it only supports
      *  files that are roughly 2GB big.)*/
-    private ArrayList<Integer> indices;
+    private ArrayList<Integer[]> indices;
```
bigerl, line 25: - [ ] Create an issue to track the limitation to 2GB. We will want to extend it at a later point. 
- [ ] The internal array should store offset + length instead of start offset + end offset (see further down)
- [ ] The internal array must be documented.

bigerl (changes requested):

iguana.corecontroller/src/test/java/org/aksw/iguana/cc/utils/FileUtilsTest.java
nck-mlcnv: will be fixed in a separate pr

iguana.corecontroller/src/main/java/org/aksw/iguana/cc/utils/IndexedLineReader.java
nck-mlcnv: I have changed the type, that the list stores, to Long

bigerl:
Please request an review from @Clueliss  when ready.

Clueliss (approved):
Hi, so I'm not sure why I'm supposed to review this, because while I do know how to write Java I wouldn't know what constitutes good Java or what is available in the stdlib.

Regarding the exceptions being rethrown or not: I don't know what the goal is here so I can't comment on what's best there.

So in that spirit, this looks fine to me and the tests seem to be passing.

iguana.corecontroller/src/main/java/org/aksw/iguana/cc/query/source/impl/FileLineQuerySource.java
```diff
@@ -18,32 +16,30 @@
...
         super(path);
-        this.queryFile = new File(this.path);
+
         try {
-            this.size = FileUtils.countLines(this.queryFile);
+            ilr = new IndexedLineReader(path);
         } catch (IOException e) {
-            LOGGER.error("Could not read queries");
+            LOGGER.error("Failed to read this file for the queries: " + path + "\n" + e);
         }
```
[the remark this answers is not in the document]
nck-mlcnv: I would suggest to fix the exception handling in a later pr, because multiple classes/methods need to be changed for this.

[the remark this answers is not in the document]
bigerl: Sounds reasonable.

iguana.corecontroller/src/main/java/org/aksw/iguana/cc/utils/FileUtils.java
```diff
@@ -14,74 +14,78 @@
...
-				for (int i = 0; i < readChars; ++i) {
-					if (c[i] == '\n') {
-						// Check if line was empty
-						if (lastChar != '\n') {
-							++count;
-						}
-					} else if (count == pos) {
-						// Now the line
-						line.append((char) c[i]);
+		try(BufferedReader br = new BufferedReader(new FileReader(file))) {
```
bigerl, line 78: Both must be managed by the try-with block. I also did that wrong in my code snippets. Otherwise, FileReader might not be closed on exceptions:
```
try (FileReader fr = new FileReader(path);
	         BufferedReader br = new BufferedReader(fr)) {
```

Please check this for all changes made in this PR and make an issue to check it throughout the code.

bigerl (changes requested):
One little thing that came to my mind. Thx @Clueliss . I should have further specified, but I think you just did what I amtes you to: have a look for logical errors in the code.

iguana.corecontroller/src/main/java/org/aksw/iguana/cc/utils/FileUtils.java
nck-mlcnv: Similar things only happen in the CLIWorker class on line 54 and ResultSizeRetriever class on line 22, if I'm not overlooking something. The ResultSizeRetriever class is actually not being used, therefore I will only create an issue for the CLIWorker

iguana.corecontroller/src/main/java/org/aksw/iguana/cc/utils/FileUtils.java
```diff
@@ -14,74 +16,79 @@
...
+	 * Counts the lines in a file efficiently. (only if the line ending is "\n") <br/>
+	 * Source: <a href="http://stackoverflow.com/a/453067/2917596">http://stackoverflow.com/a/453067/2917596</a>
 	 *
-	 * @param filename File to count lines of
-	 * @return No. of lines in File
+	 * @param filename file to count lines of
+	 * @return number of lines in the given file
 	 * @throws IOException
 	 */
 	public static int countLines(File filename) throws IOException {
```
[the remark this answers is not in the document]
nck-mlcnv: I would suggest, that we just enforce the UTF-8 encoding for files, because other parts of the code do it as well.

iguana.corecontroller/src/main/java/org/aksw/iguana/cc/utils/IndexedLineReader.java
```diff
@@ -0,0 +1,204 @@
...
+    private void indexFile() throws IOException {
+        this.indices = new ArrayList<>();
+        try(FileReader fr = new FileReader(this.file, StandardCharsets.UTF_8)) {
+            BufferedReader br = new BufferedReader(fr);
+            // The method needs to know the length of the line ending used in the file to be able to properly calculate
+            // the starting byte position of a line
+            int lineEndingLength = FileUtils.getLineEnding(this.file.getAbsolutePath()).length();
+            long index = 0;
+            String line;
+            while((line = br.readLine()) != null) {
```
bigerl, line 145: this should also be read character-wise and not line-wise. Same implementation as in indexFile(String separator) can be used.

bigerl (changes requested):
This turns out to be a much bigger change than anticipated. The general structure looks good, now. I've pointed out some bugs, smaller naming improvements and an suggestion for further testing below. 

One general comment that is also reflected in some single comments: Please use character-wise reading instead of line-wise reading wherever we read and seek files manually.

iguana.corecontroller/src/main/java/org/aksw/iguana/cc/utils/IndexedLineReader.java
```diff
@@ -0,0 +1,204 @@
...
+                if(!line.isBlank()){
+                    this.indices.add(new Long[] {index, (long) line.length()});
+                    this.size++;
+                }
+                index += line.length() + lineEndingLength;
```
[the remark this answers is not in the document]
bigerl: never mind. I was wrong

iguana.corecontroller/src/main/java/org/aksw/iguana/cc/utils/IndexedQueryReader.java
```diff
@@ -0,0 +1,165 @@
...
+        try(FileReader fr = new FileReader(this.file, StandardCharsets.UTF_8);
+            BufferedReader br = new BufferedReader(fr)) {
+            // The method needs to know the length of the line ending used in the file to be able to properly calculate
+            // the starting byte position of a line
+            long lastIndex = 0;
+            long currentIndex = 0;
+            int counter = 0;
+            int character;
+            boolean isWhiteSpace = true;
+            while((character = br.read()) != -1) {
```
bigerl, line 138: This doesn't take care of seperators like `###$`. It would miss the following: `a####$b`. (Result should actually be "a#" and "b" or (0,2), (6,1) respectively)

I would suggest something like this: 
```java
final char[] sepChars = separator.toCharArray();
            long lastQueryIndex = 0;
            long currentIndex = 0;
            int sepCounter = 0;
            int character;

            while((character = br.read()) != -1) {
                // this might become a sep counter
                boolean maybeSep = (character == (int) sepChars[sepCounter]);
                // this is a completed sep counter
                boolean isSep = maybeSep && (sepCounter + 1 == sepChars.length);

                // Before the leading characters of a separator were read but the current character does not match
                // Check if a shorter part of the tail read contains the leading characters of the separator
                if (!maybeSep && sepCounter > 0) {
                    // offset to where we started reading the separator originally
                    offset_loop:
                    for (int offset = 1; offset <= sepCounter; offset++) {
                        boolean sepSubstrMatch = true;
                        // check if separator with an offset from start is the same as from start
                        for (int i = 0; i <= sepCounter - offset; i++) {
                            if (sepChars[i] != sepChars[i + offset])
                                continue offset_loop;
                        }
                        // found a match, adjust sepCounter and maybeSep
                        sepCounter -= offset;
                        maybeSep = true;
                    }
                }
                
                if (isSep) {
                    this.indices.add(new Long[]{lastQueryIndex, currentIndex - lastQueryIndex - sepChars.length});
                    lastQueryIndex = currentIndex;
                    sepCounter = 0;
                } else if (maybeSep){
                    sepCounter++;
                }
                ++currentIndex;
            }
```

PLease doubl check and test before you use the code.

iguana.corecontroller/src/main/java/org/aksw/iguana/cc/utils/IndexedQueryReader.java
```diff
@@ -0,0 +1,165 @@
...
+        this.indices = new ArrayList<>();
+        try(FileReader fr = new FileReader(this.file, StandardCharsets.UTF_8);
+            BufferedReader br = new BufferedReader(fr)) {
+            // The method needs to know the length of the line ending used in the file to be able to properly calculate
+            // the starting byte position of a line
+            long lastIndex = 0;
+            long currentIndex = 0;
+            int counter = 0;
+            int character;
+            boolean isWhiteSpace = true;
```
bigerl, line 137: I know it was something I said that we should remove leading and trailing white spaces. It seems to make it unnecessary complicated. Did we would want them removed?

iguana.corecontroller/src/main/java/org/aksw/iguana/cc/utils/FileUtils.java
```diff
@@ -14,74 +14,82 @@
...
```
bigerl, line 24: this seems not to be used anywhere. Can be removed.

iguana.corecontroller/src/main/java/org/aksw/iguana/cc/utils/FileUtils.java
```diff
@@ -14,74 +14,82 @@
...
+	 *
+	 * @param index the position of a non-empty line which should be returned
+	 * @param file 	the file to read from
+	 * @return the line at the given position
 	 * @throws IOException
 	 */
-	public static String readLineAt(int pos, File filename) throws IOException {
-		try (InputStream is = new BufferedInputStream(new FileInputStream(filename))) {
-			StringBuilder line = new StringBuilder();
+	public static String readLineAt(int index, File file) throws IOException {
```
bigerl, line 77: Same for this function. It is not used anymore. All functionality is now in IndexedQueryReader

iguana.corecontroller/src/test/java/org/aksw/iguana/cc/query/source/impl/FileSeparatorQuerySourceTest.java
```diff
@@ -45,17 +45,17 @@ public void sizeTest() {
 
     @Test
     public void getQueryTest() throws IOException {
-        assertEquals("QUERY 1 {still query 1}", this.querySource.getQuery(0));
-        assertEquals("QUERY 2 {still query 2}", this.querySource.getQuery(1));
-        assertEquals("QUERY 3 {still query 3}", this.querySource.getQuery(2));
+        assertEquals("QUERY 1 {\r\nstill query 1\r\n}", this.querySource.getQuery(0));
```
bigerl, line 48: Please add this case: separator: "###$", string: "a####$b" Result should actually be "a#" and "b" or (0,2), (6,1) respectively

iguana.corecontroller/src/main/java/org/aksw/iguana/cc/utils/IndexedQueryReader.java
```diff
@@ -0,0 +1,165 @@
...
+     * @throws IOException
+     */
+    public String readQuery(int index) throws IOException {
+        // split long value into two integers
+        long length = this.indices.get(index)[1];
+        int upper = (int)(length >> 32);
+        int lower = (int)length;
+
+        // store content into two byte arrays (for the case that the content is >~2GB)
+        byte[] data1 = new byte[lower];
```
bigerl, line 87: In Java a string and a byte array share both the same size limit wound up by the int size type. So this approach won't work. The only option (viable) option I see to retrieve larger queries is to stream them directly from the file. For this PR I suggest to just assume for the moment that queries are smaller than 2GB :)

bigerl (changes requested):
It's getting into shape. I think we missed one case when parsing the separators. See my comments below.

bigerl (changes requested):

iguana.corecontroller/src/main/java/org/aksw/iguana/cc/utils/IndexedQueryReader.java
```diff
@@ -0,0 +1,194 @@
...
+        return this.indices.size();
+    }
+
+    /**
+     * Indexes every content in between two of the given separator. The beginning and the end of the file count as
+     * separators too.
+     * @param separator the custom separator
+     * @throws IOException
+     */
+    private void indexFile(String separator) throws IOException {
```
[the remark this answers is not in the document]
bigerl: https://docs.oracle.com/javase/8/docs/api/java/io/InputStreamReader.html

>  To enable the efficient conversion of bytes to characters, more bytes may be read ahead from the underlying stream than are necessary to satisfy the current read operation. 

OK, doesn't work. See there ↑

nck-mlcnv:
Conflicts have to be resolved and the documentation now needs checked and maybe updated now, because the behaviour of the FileSeparatorQuerySource has been changed. 

- [ ] update documentation
- [ ] resolve conflicts

bigerl:
I had a short look at some of the failing tests. They seem to be due to leading/trailing whitespaces/new line characters in tests that were removed before but are kept now.

bigerl (changes requested):
I had a short look at some of the failing tests. At least some of them seem to be caused by leading/trailing whitespaces/new line characters in tests. The expected strings do not yet have them because they were removed before by the QuerySource.

bigerl (approved):
I would remove the textfiles for the tests at some point. See #214 for that. For now, this looks fine and we can move on.

Issue #198, opened 2023-04-11
fix github actions


bigerl:
All tests on develop should be green. Also, this should run on pull requests as well: https://github.com/dice-group/IGUANA/actions/runs/4666707869/jobs/8261609422

PR #202, opened 2023-04-18
Fix unit tests

nck-mlcnv:
This pr should fix the the failing tests in the testing workflow. #198


bigerl (changes requested):
Code looks good. However, the FolderQuerySource should produce a deterministic order.

PR #206, opened 2023-05-12
Change repository structure

nck-mlcnv:
This pull request merges all maven modules to just one. The pom.xml file has also been changed accordinlgy.


bigerl (approved):
Removing the packages is fine for now. Build is successful on my system.
The structure of the packages is still not optimal but that can be tackled later.
On the high level I would for example write out cc and rp.

bigerl:
#215 super-seeded this PR. closing it.

PR #213, opened 2023-06-27
Upgrade to junit 5

nck-mlcnv:
This pr upgrades junit to junit 5. I should note that there might be some problems with running the tests in Intellij and the vintage-engine from junit 5. For some reason on my machine Intellij used junit 4.10 for the corecontroller module, which caused problems with the vintage-engine.

Actually migrating the tests to junit 5 should be done later.

### DIFF
--- a/iguana.commons/pom.xml
+++ b/iguana.commons/pom.xml
@@ -47,12 +47,6 @@
 
 	<dependencies>
 		<dependency>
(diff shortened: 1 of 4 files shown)


bigerl (changes requested):
I have rebased this one onto the current development. Please check why the tests fail.

bigerl (approved):

PR #218, opened 2023-06-30
SPARQLProtocolWorker

bigerl:
SPARQLProtocolWorker is a draft for a better, more reliable worker that is tailored towards SPARQL Protocol. Each worker uses a single HttpClient and handles work completion conditions itself. It also covers sending and receiving HTTP request and request bodies that exceed 2GB. 

This PR gives an idea what the internals of such a worker could look like. It doesn't provide a full implementation and the code is not yet used within IGUANA. 

TODOs:

- [x] Complete TODOs in SPARQLProtocolWorker
- [x] Make SPARQLProtocolWorker configurable (and remove all other workers for now)
- [ ] Implement LanguageProcessors for xml, csv and tsv SPARQL results -> move to separate issue
- [ ] Port LanguageProcessor for RDF results -> move to separate issue
- [x] Adjust Stresstest to start SPARQLProtocolWorkers
- [ ] <del>Adjust Streestest to start a ResponseBodyProcessor per unique (QuerySource,LanguageProcessor) </del>
- [ ] write/update tests for all classes touched
- [ ] document adjusted behavior
- [x] Ask @MichaelRoeder for a review of BigByteArray{Input,Output}Stream after tests are implemented for them
- [ ] Use LSQ to decide if a query is an update query; if yes, use the update endpoint -> move to separate issue

Future improvements:
- allow to select language processors for: RDF and SPARQL-RESULT for each worker. The SPARQLProtocolWorker does content type negotiation based on info from LSQ about the query (basically: if it is a construct query, expect RDF)
- Make ResponseBodyProcessor configurable, including timeout for processing after querying is done and number of threads


MichaelRoeder:
Should `BigByteArrayInputStream` and `BigByteArrayOutputStream` be thread-safe? :thinking:

nck-mlcnv:
Hi, I think there not supposed to be thread-safe.

src/main/java/org/aksw/iguana/commons/io/BigByteArrayInputStream.java
```diff
@@ -1,54 +1,101 @@
 package org.aksw.iguana.commons.io;
 
+import org.apache.hadoop.hbase.io.ByteArrayOutputStream;
+
 import java.io.IOException;
 import java.io.InputStream;
 
+import static java.lang.Math.min;
+
 public class BigByteArrayInputStream extends InputStream {
```
bigerl, line 10: It should have something to mark it as closed/done so that we can be sure it was written until the (intended) end.

src/main/java/org/aksw/iguana/cc/worker/impl/SPARQLProtocolWorker.java
```diff
@@ -0,0 +1,331 @@
...
+    private HttpExecutionResult executeHttpRequest(Duration timeout) throws IOException, URISyntaxException {
+        final QueryHandler.QueryStreamWrapper queryHandle = config().queries().getNextQueryStream();
+        final HttpRequest request = requestFactory.buildHttpRequest(
+                queryHandle.queryInputStream(),
+                timeout,
+                config().connection(),
+                config().acceptHeader()
+        );
+
+        final Instant requestStart = Instant.now();
```
bigerl, line 274: Before we take the requestStart instant, we must make sure that there is a thread in the pool. This should be done with [`pool.getActiveCount() < 2`](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ThreadPoolExecutor.html#getActiveCount--).

src/main/java/org/aksw/iguana/cc/worker/impl/SPARQLProtocolWorker.java
```diff
@@ -0,0 +1,331 @@
...
+        try {
+            return httpClient.sendAsync(request, HttpResponse.BodyHandlers.ofInputStream())
+                    .thenApply(httpResponse -> {
+                        try (final var bodyStream = httpResponse.body()) {
+                            if (httpResponse.statusCode() / 100 == 2) { // Request was successful
+                                OptionalLong contentLength = httpResponse.headers().firstValueAsLong("Content-Length");
+                                try (var hasher = hasherFactory.newStreamingHash64(0)) {
+                                    int readBytes;
+                                    while ((readBytes = bodyStream.readNBytes(this.buffer, 0, this.buffer.length)) != 0) {
+                                        hasher.update(this.buffer, 0, readBytes);
```
bigerl, line 298: Check in the look for timeouts.

bigerl (commented):
Comments from @MichaelRoeder

src/main/java/org/aksw/iguana/commons/io/BigByteArrayInputStream.java
MichaelRoeder: Some more explanation: It makes the usage of the classes safer, if it is mad explicit that the input stream cannot be used until the output stream is done (e.g., until it has been closed) and that the output stream doesn't take data after it has been closed. This avoids that the streams are used in parallel in different threads.

So my suggestion (although it can be done in a different way as well):
- in all write methods, check whether the output stream is closed. If yes, throw an exception, else write data as usual.
- in all read methods, check whether the connected output stream is closed. If yes, read as usual, else throw an exception.

src/main/java/org/aksw/iguana/cc/worker/impl/SPARQLProtocolWorker.java
nck-mlcnv: I'm assuming you're suggesting that, because timeouts after the response has been received should be ignored, right?

src/main/java/org/aksw/iguana/cc/worker/impl/SPARQLProtocolWorker.java
nck-mlcnv: Could you elaborate, which funtion you're mentioning?

src/main/java/org/aksw/iguana/cc/worker/impl/SPARQLProtocolWorker.java
bigerl: First of all: in thenApply we have the httpResponse object but the body was not yet fully received. This is a special property of the httpClient where the result is returned as InputStream BodyHandler. Consequently, we must keep our own clock for the timeout running until we have fully consumed or the timeout is reached. That's also why we should check for timeout in the while loop.

Why should we keep things it in thenApply: readNBytes is blocking. So, it waits until either the requested amount of data is available or the stream is closed. So, if a client stalls but doesn't cancel the HTTP response, readNBytes will block. This also means that the thread will go on running even if the timeout set in `.get(timeout.toNanos(), TimeUnit.NANOSECONDS);` is reached because async threads cannot be terminated unless they finish their work or the pool is shut down. SO, with the `get` method, the calling thread is waken up but the thread that was is still running. That is also why we need to put a check for an timeout into the loop reading the stream: if reached return failure or throw. 

And, that's also the reason why we must check with `pool.getActiveCount() < 2` if enough threads are available before we send the next request. Otherwise, we will wait for those threads to terminate before our workload starts.

src/main/java/org/aksw/iguana/cc/worker/impl/SPARQLProtocolWorker.java
bigerl: see below

src/main/java/org/aksw/iguana/cc/worker/impl/SPARQLProtocolWorker.java
nck-mlcnv: I gave the httpclient in the sparqlworker a seperate ExecutorService. With that it's possible to forcefully shutdown a running task before the next request is sent.

nck-mlcnv:
Just as note, some tests don't seem to be running for some reason.

PR #219, opened 2023-07-06
Rework result processing

nck-mlcnv:
This PR mainy aims to renew the result processing as it was quite messy. Changes:

- Properties class isn't used anymore, which was quite unwieldy
- most metrics don't write RDF models anymore and only calculate single number values
- added usage of static RDF objects
- penalized metrics are now their own metrics, penalty needs to be given in the configuration now
- "new" metric "Aggregated Executions Statistics"
- the Stresstest class is now managing the result processing itself
- changed repository structure


src/main/java/org/aksw/iguana/cc/config/IguanaConfig.java
```diff
@@ -127,42 +127,46 @@ private RPController initResultProcessor() {
 			LOGGER.info("No metrics were set. Using default metrics.");
 			metrics = new ArrayList<>();
 			MetricConfig config = new MetricConfig();
-			config.setClassName(QMPHMetric.class.getCanonicalName());
+			config.setClassName(QMPH.class.getCanonicalName());
```
bigerl, line 130: I would suggest to drop the configurability of metrics altogether.

bigerl (changes requested):
In general it looks good. I see a lot of things that will change with the next open PR #218 is finished. So I tend towards merging this one fast into a new branch major-rewrite/develop that we branch from develop and continue then with #218.

src/main/java/org/aksw/iguana/cc/tasks/stresstest/storage/impl/CSVStorage.java
```diff
@@ -0,0 +1,294 @@
...
+     * Creates a CSV file with the given name values that will be located inside the parent folder. The name value are
+     * joined together with the character '-'. Empty values will be ignored.
+     *
+     * @param nameValues strings that build up the name of the file
+     * @throws IOException if an I/O error occurs
+     * @return path object to the created CSV file
+     */
+    private Path createCSVFile(String... nameValues) throws IOException {
+        // remove empty string values
+        nameValues = Arrays.stream(nameValues).filter(Predicate.not(String::isEmpty)).toArray(String[]::new);
```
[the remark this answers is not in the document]
nck-mlcnv: The connection version value is optional. I set the default value for it as an empty String in the setObjectAttibutes method.

src/main/java/org/aksw/iguana/cc/tasks/stresstest/storage/impl/CSVStorage.java
```diff
@@ -0,0 +1,294 @@
...
+     * This method sets the objects attributes by querying the given model.
+     *
+     * @param data the result model
+     * @throws NoSuchElementException might be thrown if the model is incorrect
+     */
+    private void setObjectAttributes(Model data) throws NoSuchElementException {
+        ResIterator resIterator = data.listSubjectsWithProperty(RDF.type, IONT.dataset);
+        Resource datasetRes = resIterator.nextResource();
+        NodeIterator nodeIterator = data.listObjectsOfProperty(datasetRes, RDFS.label);
+        this.dataset = nodeIterator.next().asLiteral().getLexicalForm();
```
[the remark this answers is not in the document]
nck-mlcnv: The StresstestResultProcessor calls the interface funtion `storeResults` at the end of a task and sends the result model only for that current task. So the model that the interface funtion receives has contains only the dataset of the executed task

src/main/java/org/aksw/iguana/cc/tasks/stresstest/storage/impl/CSVStorage.java
```diff
@@ -0,0 +1,294 @@
...
+
+        resIterator = data.listSubjectsWithProperty(RDF.type, IONT.connection);
+        Resource connectionRes = resIterator.nextResource();
+        nodeIterator = data.listObjectsOfProperty(connectionRes, RDFS.label);
+        this.connection = nodeIterator.next().asLiteral().getLexicalForm();
```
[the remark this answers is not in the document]
nck-mlcnv: Its the same as the dataset

src/main/java/org/aksw/iguana/cc/lang/impl/SPARQLLanguageProcessor.java
```diff
@@ -70,23 +70,24 @@ public Model generateTripleStats(List<QueryWrapper> queries, String resourcePref
         Model model = ModelFactory.createDefaultModel();
         for(QueryWrapper wrappedQuery : queries) {
             Resource subject = ResourceFactory.createResource(COMMON.RES_BASE_URI + resourcePrefix + "/" + wrappedQuery.getId());
-            model.add(subject, RDF.type, Vocab.queryClass);
-            model.add(subject, Vocab.rdfsID, wrappedQuery.getId().replace("sparql", ""));
+            model.add(subject, RDF.type, IONT.query);
+            // TODO: queryID is already used in a different context
```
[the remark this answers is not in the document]
nck-mlcnv: queryID is already used to assign an ExecutedQuery to a Query (see results schema)

src/main/java/org/aksw/iguana/cc/config/IguanaConfig.java
bigerl: for later

src/main/java/org/aksw/iguana/cc/worker/AbstractWorker.java
```diff
@@ -184,19 +175,9 @@ protected HttpContext getAuthContext(String endpoint) {
     }
 
     public synchronized void addResults(QueryExecutionStats results) {
+        // TODO: check if statement for bugs, if the if line exists in the UpdateWorker, the UpdateWorker fails its tests
```
bigerl, line 178: This one can be ignored for now because the UpdateWorker will be rewritten anyway.

src/test/java/org/aksw/iguana/cc/tasks/storage/impl/CSVStorageTest.java
```diff
@@ -0,0 +1,201 @@
...
+        Resource workerRes2 = IRES.getResource("worker2");
+        Resource taskQueryRes = IRES.getResource("task-query");
+        Resource workerQueryRes1 = IRES.getResource("worker-query-1");
+        Resource workerQueryRes2 = IRES.getResource("worker-query-2");
+        Resource queryRes1 = IRES.getResource("query1");
+        Resource queryRes2 = IRES.getResource("query2");
+
+        Model m = ModelFactory.createDefaultModel();
+
+        m.add(experimentRes, RDF.type, IONT.experiment);
```
bigerl, line 69: use method cascading with add

src/test/java/org/aksw/iguana/cc/tasks/storage/impl/CSVStorageTest.java
```diff
@@ -0,0 +1,201 @@
...
+    Path folder;
+    Path suiteFolder;
+
+    @BeforeEach
+    public void setup() throws IOException {
+        this.folder = Files.createTempDirectory("iguana-CSVStorage-test");
+        this.suiteFolder = folder.resolve(IguanaConfig.getSuiteID());
+    }
+
+    public static Arguments createTestData1() {
```
bigerl, line 50: The data generation of this test is highly fragile in a sense that small changes to the rdf serialization render the generated data incorrec.t. It could be more robust if it relied on StresstestResultProcessor directly to generate a jena model. 
This applies in a similar fashion to the other tests in this package as well. This doesn't need to be patched but can be moved to an issue.

bigerl (approved):
Looks good. There is nothing in the comments that should still be resolved in this PR. So, we are good to go to merge this.

PR #221, opened 2023-10-06
HTTP worker refactoring

nck-mlcnv:
SPARQLProtocolWorker is a draft for a better, more reliable worker that is tailored towards SPARQL Protocol. Each worker uses a single HttpClient and handles work completion conditions itself. <del>It also covers sending and receiving HTTP request and request bodies that exceed 2GB. </del>

This PR gives an idea what the internals of such a worker could look like. It doesn't provide a full implementation and the code is not yet used within IGUANA. 

TODOs:

- [x] Complete TODOs in SPARQLProtocolWorker
- [x] Make SPARQLProtocolWorker configurable (and remove all other workers for now)
- [x] <del>Implement LanguageProcessors for xml, csv and tsv SPARQL results -> move to separate issue</del>
- [x] <del>Port LanguageProcessor for RDF results -> move to separate issue</del>
- [x] Adjust Stresstest to start SPARQLProtocolWorkers
- [x] <del>Adjust Streestest to start a ResponseBodyProcessor per unique (QuerySource,LanguageProcessor) </del>
- [x] write/update tests for all classes touched
- [x] <del>document adjusted behavior</del>
- [x] Ask @MichaelRoeder for a review of BigByteArray{Input,Output}Stream after tests are implemented for them
- [x] <del>Use LSQ to decide if a query is an update query; if yes, use the update endpoint -> move to separate issue</del>

Future improvements:
- allow to select language processors for: RDF and SPARQL-RESULT for each worker. The SPARQLProtocolWorker does content type negotiation based on info from LSQ about the query (basically: if it is a construct query, expect RDF)
- Make ResponseBodyProcessor configurable, including timeout for processing after querying is done and number of threads


src/main/java/org/aksw/iguana/cc/worker/impl/SPARQLProtocolWorker.java
```diff
@@ -0,0 +1,412 @@
...
+                    if (!used) {
+                        used = true;
+                        return () -> queryStream;
+                    }
+                    else
+                        return () -> null;
+                }
+            }
+
+            final var streamSupplier = new CustomStreamSupplier();
```
bigerl, line 85: instantiate only in the cases where it is actually used

src/main/java/org/aksw/iguana/cc/worker/impl/SPARQLProtocolWorker.java
```diff
@@ -0,0 +1,412 @@
...
+                                    .setParameter("query",
+                                            new String(queryStream.readAllBytes(), StandardCharsets.UTF_8))
+                                    .build())
+                            .GET();
+                }
+                case POST_URL_ENC_QUERY -> {
+                    request.uri(connection.endpoint())
+                            .header("Content-Type", "application/x-www-form-urlencoded")
+                            .POST(HttpRequest.BodyPublishers.ofString(
+                                    urlEncode(Collections.singletonList(
```
bigerl, line 105: (1): create an issue to either url-encode on the fly or pre-encode the query string and supply it encoded as stream

src/main/java/org/aksw/iguana/cc/worker/impl/SPARQLProtocolWorker.java
```diff
@@ -0,0 +1,412 @@
...
+                case POST_QUERY -> {
+                    request.uri(connection.endpoint())
+                            .header("Content-Type", "application/sparql-query")
+                            .POST(HttpRequest.BodyPublishers.ofInputStream(streamSupplier.getStreamSupplier()));
+                }
+                case POST_URL_ENC_UPDATE -> {
+                    request.uri(connection.endpoint())
+                            .header("Content-Type", "application/x-www-form-urlencoded")
+                            .POST(HttpRequest.BodyPublishers.ofString(
+                                    urlEncode(Collections.singletonList(
```
bigerl, line 118: see (1)

src/main/java/org/aksw/iguana/cc/worker/impl/SPARQLProtocolWorker.java
```diff
@@ -0,0 +1,412 @@
...
+            OptionalLong hash,
+            Optional<Exception> exception
+    ) {
+        public boolean completed() {
+            return response.isPresent();
+        }
+
+        public boolean successful() {
+            if (completed() && exception.isEmpty())
+                return (response.get().statusCode() / 100) == 2;
```
bigerl, line 178: Verify that all unchecked optional accesses are safe. Please do research if there is any way to indicate that the check was done by other means.

src/main/java/org/aksw/iguana/cc/worker/impl/SPARQLProtocolWorker.java
```diff
@@ -0,0 +1,412 @@
...
+
+
+    public CompletableFuture<Result> start() {
+        return CompletableFuture.supplyAsync(() -> {
+            List<ExecutionStats> executionStats = new ArrayList<>();
+            if (config().completionTarget() instanceof QueryMixes queryMixes) {
+                for (int i = 0; i < queryMixes.number(); i++) {
+                    for (int j = 0; j < config().queries().getQueryCount(); j++) {
+                        ExecutionStats execution = executeQuery(config().timeout(), false);
+                        if (execution != null) {
```
bigerl, line 222: execution should never be null here, right? better use an assertion then. Or even throw if null

src/main/java/org/aksw/iguana/cc/worker/impl/SPARQLProtocolWorker.java
```diff
@@ -0,0 +1,412 @@
...
+            } else if (config().completionTarget() instanceof TimeLimit timeLimit) {
+                final Instant endTime = Instant.now().plus(timeLimit.duration());
+                Instant now;
+                while ((now = Instant.now()).isBefore(endTime)) {
+                    final Duration timeToEnd = Duration.between(now, endTime);
+                    final boolean timeoutBeforeEnd = config().timeout().compareTo(timeToEnd) < 0;
+                    final Duration thisQueryTimeOut = (timeoutBeforeEnd) ? config().timeout() : timeToEnd;
+                    // If timeoutBeforeEnd is false, fail shouldn't be counted as timeout
+                    ExecutionStats execution = executeQuery(thisQueryTimeOut, !timeoutBeforeEnd);
+                    if (execution != null){
```
bigerl, line 238: comment for if

src/main/java/org/aksw/iguana/cc/worker/impl/SPARQLProtocolWorker.java
```diff
@@ -0,0 +1,412 @@
...
+    public SPARQLProtocolWorker(long workerId, ResponseBodyProcessor responseBodyProcessor, Config config) {
+        super(workerId, responseBodyProcessor, config);
+        this.responseBodyProcessor = responseBodyProcessor;
+        this.executor = (ThreadPoolExecutor) Executors.newFixedThreadPool(1);
+        this.requestFactory = new RequestFactory(config().requestType());
+        this.httpClient = buildHttpClient();
+    }
+
+
+    public CompletableFuture<Result> start() {
```
bigerl, line 215: add docstring

src/main/java/org/aksw/iguana/cc/worker/impl/SPARQLProtocolWorker.java
```diff
@@ -0,0 +1,412 @@
...
+                    Duration.ZERO,
+                    Optional.empty(),
+                    OptionalLong.empty(),
+                    OptionalLong.empty(),
+                    Optional.of(e)
+            );
+        }
+
+        if (((ThreadPoolExecutor) this.httpClient.executor().get()).getActiveCount() != 0) {
+            ((ThreadPoolExecutor) this.httpClient.executor().get()).shutdownNow();
```
bigerl, line 327: not tested this but I would guess that if you shutdown the executor you need to create a new one. Also, it doesn't guarantee to shutdown the running tasks: "There are no guarantees beyond best-effort attempts to stop processing actively executing tasks.". so, it would be better to wait and write the waiting time to the log. Log should contain: thread id, time when waiting started, current time. Sad but true: we can't catch tasks gone rouge.

bigerl (changes requested):
Partial review. Only worked my way up through source not tests to SPARQLProtocol worker yet. Looks good in general so far. Find some comments below.

nck-mlcnv:
Hi, thanks for the review, I will look into the change requests soon.

src/main/java/org/aksw/iguana/cc/worker/HttpWorker.java
```diff
@@ -0,0 +1,165 @@
...
+            int queryID,
+            Instant startTime,
+            Duration duration, // should always exist
+            Optional<Integer> httpStatusCode,
+            OptionalLong contentLength,
+            OptionalLong responseBodyHash,
+            Optional<Exception> error
+    ) {
+        public enum END_STATE {
+            SUCCESS(1), // TODO: values are the same as previous implementation for backwards compatibility, I think descriptive String values would be better
```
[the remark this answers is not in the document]
nck-mlcnv: I just checked them and changed what I could find. But I'm not sure though, if they're correct. Also maybe we should add more error codes in the future.

src/main/java/org/aksw/iguana/cc/worker/impl/SPARQLProtocolWorker.java
nck-mlcnv: I don't know exactly what you meant with this comment, but I'm just going to assume, that you meant that I should move the comment to the if statement. 😅

src/main/java/org/aksw/iguana/cc/worker/impl/SPARQLProtocolWorker.java
nck-mlcnv: Won't there be a new executor created with the buildHttpClient() method?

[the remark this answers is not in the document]
nck-mlcnv: The method shutdownNow() is supposed to call the interrupt method of all it's threads.

nck-mlcnv: Also, the shutdownNow() method is supposed to call interrupt() on all the threads it has. And if I'm not mistaken the http response input stream is supposed to respond to the interrupt flag (it looks like there was acutally a [bug ](https://bugs.openjdk.org/browse/JDK-8294047) related to this). So maybe this should actually never block indefinetely?

bigerl: The bug report mentions that the bug is closed only in openjdk 20. We should document that somewhere.

src/main/java/org/aksw/iguana/cc/storage/Storable.java
```diff
@@ -0,0 +1,36 @@
...
+public interface Storable {
+
+    record CSVData (
+            String folderName,
+            List<CSVFileData> files
+    ) {
+        public record CSVFileData(String filename, String[][] data) {}
+    }
+
+    interface AsCSV {
```
bigerl, line 16: toCSV is not used by CSVStorage. So, what is it good for?

src/main/java/org/aksw/iguana/cc/suite/IguanaSuiteParser.java
```diff
@@ -0,0 +1,246 @@
...
+                    final var datasetName = node.asText();
+                    if (!datasets.containsKey(datasetName))
+                        throw new IllegalStateException(MessageFormat.format("Unknown dataset name: {0}", datasetName));
+                    return datasets.get(datasetName);
+                } else {
+                    DatasetConfig datasetConfig = ctxt.readValue(jp, DatasetConfig.class);
+                    if (datasets.containsKey(datasetConfig.name()))
+                        assert datasets.get(datasetConfig.name()) == datasetConfig;
+                    else datasets.put(datasetConfig.name(), datasetConfig);
+                    return datasetConfig; // TODO: double check if this really works
```
bigerl, line 121: resolve TODO

src/main/java/org/aksw/iguana/cc/suite/IguanaSuiteParser.java
```diff
@@ -0,0 +1,246 @@
...
+                    final var connectionName = node.asText();
+                    if (!connections.containsKey(connectionName))
+                        throw new IllegalStateException(MessageFormat.format("Unknown connection name: {0}", connectionName));
+                    return connections.get(connectionName);
+                } else {
+                    ConnectionConfig connectionConfig = ctxt.readValue(jp, ConnectionConfig.class);
+                    if (connections.containsKey(connectionConfig.name()))
+                        assert connections.get(connectionConfig.name()) == connectionConfig;
+                    else connections.put(connectionConfig.name(), connectionConfig);
+                    return connectionConfig; // TODO: double check if this really works
```
bigerl, line 153: resolve TODO

src/main/java/org/aksw/iguana/cc/suite/IguanaSuiteParser.java
```diff
@@ -0,0 +1,246 @@
...
+                return Duration.parse(durationString);
+            }
+        }
+
+        mapper = new ObjectMapper(factory).registerModule(new JavaTimeModule())
+                .registerModule(new SimpleModule()
+                        .addDeserializer(DatasetConfig.class, new DatasetDeserializer())
+                        .addDeserializer(ConnectionConfig.class, new ConnectionDeserializer())
+                        .addDeserializer(Duration.class, new HumanReadableDurationDeserializer()));
+        // TODO: update validator and reactivate
```
bigerl, line 197: resolve TODO

src/main/java/org/aksw/iguana/cc/suite/IguanaSuiteParser.java
```diff
@@ -0,0 +1,246 @@
...
+//        if(validate && !validateConfig(config, schemaFile, mapper)){
+//            return null;
+//        }
+        // TODO: use date or something random as source for the suite ID
```
bigerl, line 201: resolve TODO

src/main/java/org/aksw/iguana/cc/suite/IguanaSuiteParser.java
```diff
@@ -0,0 +1,246 @@
...
+    private static Map<String, ConnectionConfig> preparseConnections(ObjectMapper mapper, String input) throws JsonProcessingException {
+        @JsonIgnoreProperties(ignoreUnknown = true)
+        record PreparsingConnections(@JsonProperty(required = true) List<ConnectionConfig> connections) {}
+        final var preparsingConnections = mapper.readValue(input, PreparsingConnections.class);
+
+        return preparsingConnections.connections().stream().collect(Collectors.toMap(ConnectionConfig::name, Function.identity()));
+    }
+
+    private static boolean validateConfig(Path config, String schemaFile, ObjectMapper mapper) throws IOException {
+        // TODO: update
```
bigerl, line 231: resolve TODO

src/main/java/org/aksw/iguana/cc/worker/HttpWorker.java
```diff
@@ -0,0 +1,163 @@
...
+public abstract class HttpWorker {
+
+    @JsonTypeInfo(use = JsonTypeInfo.Id.NAME,
+            include = JsonTypeInfo.As.PROPERTY,
+            property = "type")
+    @JsonSubTypes({
+            @JsonSubTypes.Type(value = SPARQLProtocolWorker.Config.class, name = "SPARQLProtocolWorker"),
+    })
+    public interface Config {
+        // TODO: add delay or complete after
```
bigerl, line 36: move TODO to issue

src/main/java/org/aksw/iguana/cc/worker/HttpWorker.java
```diff
@@ -0,0 +1,163 @@
...
+ */
+public abstract class HttpWorker {
+
+    @JsonTypeInfo(use = JsonTypeInfo.Id.NAME,
+            include = JsonTypeInfo.As.PROPERTY,
+            property = "type")
+    @JsonSubTypes({
+            @JsonSubTypes.Type(value = SPARQLProtocolWorker.Config.class, name = "SPARQLProtocolWorker"),
+    })
+    public interface Config {
```
bigerl, line 35: add start time and end time (like in Stresstest)

src/main/java/org/aksw/iguana/cc/tasks/impl/Stresstest.java
```diff
@@ -0,0 +1,111 @@
...
+        srp.process(results.workerResults);
+        srp.calculateAndSaveMetrics(results.startTime, results.endTime);
+    }
+
+    private Result executeWorkers(List<HttpWorker> workers) {
+        List<HttpWorker.Result> results = new ArrayList<>(workers.size());
+        Calendar startTime = Calendar.getInstance();
+        var futures = workers.stream().map(HttpWorker::start).toList();
+        CompletableFuture.allOf(futures.toArray(new CompletableFuture[0])).join();
+        Calendar endTime = Calendar.getInstance(); // TODO: add start and end time for each worker
```
bigerl, line 95: remove TODO

src/main/java/org/aksw/iguana/cc/tasks/impl/Stresstest.java
```diff
@@ -0,0 +1,111 @@
...
+
+    public record Config(
+            List<HttpWorker.Config> warmupWorkers,
+            @JsonProperty(required = true) List<HttpWorker.Config> workers
+    ) implements Task.Config {}
+
+    public record Result(
+            List<HttpWorker.Result> workerResults,
+            Calendar startTime,
+            Calendar endTime
```
bigerl, line 31: move startTime and endTime to Task. Might need to change it to an abstract class for that.

src/main/java/org/aksw/iguana/cc/worker/HttpWorker.java
```diff
@@ -0,0 +1,163 @@
...
+import java.util.List;
+import java.util.Optional;
+import java.util.OptionalLong;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
+
+/**
+ * Interface for the Worker Thread used in the {@link Stresstest}
+ *
+ * @author f.conrads
```
bigerl, line 25: remove author tag

src/main/java/org/aksw/iguana/commons/rdf/IONT.java
```diff
@@ -20,11 +19,7 @@ public class IONT {
     public static final Resource metric = ResourceFactory.createResource(NS + "Metric");
 
     public static Resource getMetricClass(Metric metric) {
-        // TODO: compare with stresstest class
+        // TODO: compare with stresstest class (stresstest class as a subclass of Task is iont:Stresstest while QPS for example is iont:metric/QPS)
```
bigerl, line 22: I don't get this comment. Hope it's not originally from me :)

src/test/java/org/aksw/iguana/cc/config/elements/ConnectionConfigTest.java
```diff
@@ -0,0 +1,107 @@
...
+                        """
+                )
+        );
+
+    }
+
+    @ParameterizedTest
+    @MethodSource("testData")
+    @Disabled("Doesn't work with testing default values and we don't serialize the configurations, so it's not really needed.")
+    public void testSerialization(ConnectionConfig config, String expectedJson) throws Exception {
```
bigerl, line 96: It might make sense to generate configs for users but then probably better directly in yaml. What exactly is the problem with the generated serializations?

src/test/java/org/aksw/iguana/commons/io/BigByteArrayOutputStreamTest.java
```diff
@@ -0,0 +1,310 @@
...
+import java.nio.ByteBuffer;
+import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Random;
+import java.util.function.Supplier;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+@Disabled("Big test")
```
bigerl, line 21: why is this one disabled?

src/test/java/org/aksw/iguana/cc/suite/IguanaSuiteParserTest.java
```diff
@@ -0,0 +1,17 @@
+package org.aksw.iguana.cc.suite;
+
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.Test;
+
+import java.nio.file.Path;
+
+@Disabled("In progress")
```
bigerl, line 9: TODO: implement

src/test/java/org/aksw/iguana/commons/io/BigByteArrayInputStreamTest.java
```diff
@@ -0,0 +1,224 @@
...
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Random;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+@Disabled("Big test")
```
bigerl, line 17: TODO: why is this one disabled?

src/test/java/org/aksw/iguana/cc/utils/CLIProcessManagerTest.java
```diff
@@ -1,13 +1,13 @@
...
-import org.junit.Test;
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.Test;
 
 import java.io.IOException;
 
 import static org.junit.Assert.*;
 
-@Ignore("CLI doesn't work right now")
+@Disabled("CLI doesn't work right now")
```
bigerl, line 10: remov the file

src/test/java/org/aksw/iguana/cc/storage/impl/TriplestoreStorageTest.java
```diff
@@ -26,12 +20,13 @@
  * @author f.conrads
  *
  */
+@Disabled("TriplestoreStorage hasn't been looked at in a while and the test is outdated.")
```
bigerl, line 23: should be re-enabled. For testing, use both TriplestoreStorage and RDFStorage and check that both contain the same graph. 

Query to retrieve the whole graph:
```SPARQL
construct {?x ?y ?z } where {?x ?y ?z}
```

There might also be a even simpler way with jena. 

We produce not blank nodes, right? Then checking for equality should be trivial. Otherwise, the models/graphs should be isomorphic. jena has methods for that.

bigerl (changes requested):
review covering some more of the code. I did only a superficially check on the tests until now because there seemed to be open parts.

src/test/java/org/aksw/iguana/commons/io/BigByteArrayInputStreamTest.java
nck-mlcnv: I'm not sure which of the BigByteArrayStreams it was, but one of the tests takes a long time (more than 2 minutes on my machine). I didn't want to run them each time I wanted to run all the tests, especially considering that I'm not touching these classes anymore. I will reable them soon, to test if they run on the GitHub machines.

src/test/java/org/aksw/iguana/commons/io/BigByteArrayOutputStreamTest.java
nck-mlcnv: explained in other comment

src/main/java/org/aksw/iguana/cc/storage/Storable.java
nck-mlcnv: It is used [here](https://github.com/dice-group/IGUANA/blob/4f0a309a4b56e8059488ec16da2080fc42fbc21f/src/main/java/org/aksw/iguana/cc/storage/impl/CSVStorage.java#L192C19-L192C19)

src/main/java/org/aksw/iguana/cc/suite/IguanaSuiteParser.java
nck-mlcnv: Should work 👍

src/main/java/org/aksw/iguana/commons/rdf/IONT.java
nck-mlcnv: The inherited metric classes like QPS have the uri `http://iguana-benchmark.eu/class/metric/QPS`, while the stresstest, which inherits from the class task, only has the uri `http://iguana-benchmark.eu/class/Stresstest`. The difference is, that the metric class has a `metric/`-"prefix". That was what I wanted to point out. Maybe we should give the stresstest a `task/`-prefix as well?

src/test/java/org/aksw/iguana/cc/config/elements/ConnectionConfigTest.java
nck-mlcnv: The problem was, that it needed a different testing set. The current and only test set for the connection config has json values that leave out properties of the ConnectionConfig, because they have default values. If you create a new ConnectionConfig, it instantiates those values and the mapper writes them out. This leads to the failure of the assertions.

src/test/java/org/aksw/iguana/cc/storage/impl/TriplestoreStorageTest.java
nck-mlcnv: I did a little bit differnt now. I'm basically just checking, if the sent contains the correct data.

src/test/java/org/aksw/iguana/commons/io/BigByteArrayInputStreamTest.java
bigerl: Let's keep this open then until that is clarified.

src/main/java/org/aksw/iguana/commons/rdf/IONT.java
bigerl: What do we actually use the metric classes for? Shouldn't the properties be enough?

src/main/java/org/aksw/iguana/cc/lang/LanguageProcessor.java
```diff
@@ -1,55 +1,62 @@
...
+    }
+
+    public abstract LanguageProcessingData process(InputStream inputStream, long hash);
+
+    final private static Map<String, Class<? extends LanguageProcessor>> processors = new HashMap<>();
+
+    static {
+        processors.put("application/sparql-results+json", SaxSparqlJsonResultCountingParser.class);
+        // TODO: find better solution, because the code underneath doesn't work
+//        for (LanguageProcessor processor : ServiceLoader.load(LanguageProcessor.class)) {
```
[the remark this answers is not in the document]
nck-mlcnv: We're using parts of the spring-framework now to do this.

nck-mlcnv:
I accidently enabled the BigByteArrayStream tests, and it looks like they cause some issues on the GitHub system. 🤔

src/main/java/org/aksw/iguana/commons/rdf/IONT.java
nck-mlcnv: The metric classes give the metric properties their type. 
grafik

src/test/java/org/aksw/iguana/commons/io/BigByteArrayInputStreamTest.java
bigerl: doesn't run on GitHubActions

src/main/java/org/aksw/iguana/cc/suite/IguanaSuiteParser.java
```diff
@@ -0,0 +1,260 @@
...
+            }
+        }
+
+        mapper = new ObjectMapper(factory).registerModule(new JavaTimeModule())
+                .registerModule(new SimpleModule()
+                        .addDeserializer(DatasetConfig.class, new DatasetDeserializer())
+                        .addDeserializer(ConnectionConfig.class, new ConnectionDeserializer())
+                        .addDeserializer(Duration.class, new HumanReadableDurationDeserializer()));
+
+        final String suiteID = Instant.now().getEpochSecond() + "-" + Integer.toUnsignedString(input.hashCode()); // convert to unsigned, so that there is no double -- minus in the string
```
bigerl, line 232: remind me: Instant.now() was wall and non-monotonic time, right? would make to replace them all.

bigerl (changes requested):
Some minor things.

src/main/java/org/aksw/iguana/cc/metrics/ModelWritingMetric.java
```diff
@@ -5,16 +5,16 @@
...
 import org.apache.jena.rdf.model.ModelFactory;
 
-import javax.annotation.Nonnull;
+import javax.validation.constraints.NotNull;
 import java.util.List;
 import java.util.Map;
 
 public interface ModelWritingMetric {
-    default @Nonnull Model createMetricModel(List<HttpWorker> workers, List<HttpWorker.ExecutionStats>[][] data, IRES.Factory iresFactory) {
+    default @NotNull Model createMetricModel(List<HttpWorker> workers, List<HttpWorker.ExecutionStats>[][] data, IRES.Factory iresFactory) {
```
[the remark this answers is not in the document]
nck-mlcnv: Should I remove them?

src/main/java/org/aksw/iguana/cc/suite/IguanaSuiteParser.java
nck-mlcnv: Yes I think so. But I don't think `Instant.now()` should be an issue here, because it's not used here to measure elapsed time, it's only used as a timestamp. On the other hand `System.nanoTime()` counts the time from an arbitrary point of time, so if it's used as a "timestamp", its value would be quite meaningless other than being some random numbers: https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/System.html#nanoTime()

nck-mlcnv: But I can change it if you want

bigerl (approved):

PR #224, opened 2023-10-24
Fix the calculation of the penalized metrics

nck-mlcnv:
The penalized metrics were only calculated with the successful executions, instead of every execution.


bigerl (approved):
Looks good. My changes don't change anything in the behaviour but ar solely to improve readablility. It helped me to understand what happens there so it might also help whomever might ever try to understand this piece of code.

PR #236, opened 2023-11-20
SPARQLLanguageProcessor: fix getContentTypeVal

nomoa:
Use the name of the first HeaderElement to extract the actual content-type.
This is similar to what is done by ContentType::get(HttpEntity).


nomoa:
v3 no longer maintained

PR #238, opened 2023-11-20
Fix MetricsTest

nomoa:
First commit adds a new main method to the MetricsTest test class that helps to regenerate the test resources. It also update the existing test resources to re-order them.
The second commit actually update the fixtures in accordance to the fix made in aee83ef.


nck-mlcnv:
First of all, I would like to thank you for your contributions to IGUANA.
Unfortunately, they come at a somewhat unfortunate time, as we are about to release the new v4 of IGUANA. This is a major overhaul of large parts of the code base. Since we only have limited resources for IGUANA, we can no longer maintain v3.
The v4 is in the final test phase before it is merged into the main branch. You can try it out in on the develop branch. Please keep in mind that the documentation is still being updated ATM.

nomoa:
Sure, I understand.

PR #241, opened 2023-11-28
Fix performance issues

nck-mlcnv:
During a test of the new iguana version, I noticed that there were some severe performance issues during the processing of the results, after the stresstest has finished. The biggest issue of it was the jena library and the `canonical` method inside its `XSDDuration` class, which didn't support nanoseconds as decimal digits for the durations. 
This PR aims to circumvent this issue by creating our own LiteralType class, which might not be the best solution.

I've also made the QuerySource cache its hashCode, which also caused some performance issues, as the hashCode is needed everytime the id of a Query from a QueryHandler is retrieved. Caching the QuerySource hashCode shouldn't be a problem though, as we don't expect the hashCode to change itself during the benchmark.

### DIFF
--- a/src/main/java/org/aksw/iguana/cc/metrics/impl/AggregatedExecutionStatistics.java
+++ b/src/main/java/org/aksw/iguana/cc/metrics/impl/AggregatedExecutionStatistics.java
@@ -6,6 +6,7 @@ import org.aksw.iguana.cc.worker.HttpWorker;
 import org.aksw.iguana.commons.rdf.IONT;
 import org.aksw.iguana.commons.rdf.IPROP;
 import org.aksw.iguana.commons.rdf.IRES;
+import org.aksw.iguana.commons.time.TimeUtils;
 import org.apache.jena.rdf.model.Model;
 import org.apache.jena.rdf.model.ModelFactory;
 import org.apache.jena.rdf.model.Resource;
@@ -17,9 +18,6 @@ import java.time.Duration;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
-
-import static org.aksw.iguana.commons.time.TimeUtils.createTypedDurationLiteral;
-import static org.aksw.iguana.commons.time.TimeUtils.toXSDDurationInSeconds;
 
 public class AggregatedExecutionStatistics extends Metric implements ModelWritingMetric {
 
@@ -81,8 +79,8 @@ public class AggregatedExecutionStatistics extends Metric implements ModelWritin
         m.add(queryRes, IPROP.timeOuts, ResourceFactory.createTypedLiteral(timeOuts));
         m.add(queryRes, IPROP.wrongCodes, ResourceFactory.createTypedLiteral(wrongCodes));
         m.add(queryRes, IPROP.unknownException, ResourceFactory.createTypedLiteral(unknownExceptions));
-        // m.add(queryRes, IPROP.totalTime, ResourceFactory.createTypedLiteral(toXSDDurationInSeconds(totalTime)));
-        m.add(queryRes, IPROP.totalTime, createTypedDurationLiteral(totalTime));
+        m.add(queryRes, IPROP.totalTime, TimeUtils.createTypedDurationLiteralInSeconds(totalTime));
+        // m.add(queryRes, IPROP.totalTime, TimeUtils.createTypedDurationLiteral(totalTime));
         m.add(queryRes, RDF.type, IONT.executedQuery);
 
         return m;
--- a/src/main/java/org/aksw/iguana/cc/metrics/impl/AggregatedExecutionStatistics.java
+++ b/src/main/java/org/aksw/iguana/cc/metrics/impl/AggregatedExecutionStatistics.java
@@ -18,6 +18,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 
+import static org.aksw.iguana.commons.time.TimeUtils.createTypedDurationLiteral;
 import static org.aksw.iguana.commons.time.TimeUtils.toXSDDurationInSeconds;
 
 public class AggregatedExecutionStatistics extends Metric implements ModelWritingMetric {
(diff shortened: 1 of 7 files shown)


src/main/java/org/aksw/iguana/commons/time/TimeUtils.java
```diff
@@ -23,8 +22,23 @@ public static XSDDuration toXSDDurationInSeconds(Duration duration) {
 		return (XSDDuration) new XSDDurationType().parse("PT" + new BigDecimal(BigInteger.valueOf(duration.toNanos()), 9).toPlainString() + "S");
 	}
 
+	public static Literal createTypedDurationLiteralInSeconds(Duration duration) {
+		var seconds = "PT" + new BigDecimal(BigInteger.valueOf(duration.toNanos()), 9).toPlainString() + "S";
```
bigerl, line 26: Is it really necessary to have the seconds string here? Otherwise you could just use `duration.toString()` which also generates valid strings for XSD duration (i.e. ISO-8601).

src/test/java/org/aksw/iguana/cc/query/list/QueryListTest.java
```diff
@@ -100,9 +101,13 @@ public static List<Arguments> data() throws IOException {
         return out;
     }
 
+    @Test
     public void testIllegalArguments() {
         assertThrows(NullPointerException.class, () -> new InMemQueryList(null));
-        assertThrows(NullPointerException.class, () -> new FileBasedQueryList(null));
+        assertThrows(NullPointerException.class, () -> {
+            final var ql = new FileBasedQueryList(null);
+            ql.size();
```
bigerl, line 109: what is the ql.size() call good for?

src/main/java/org/aksw/iguana/cc/metrics/impl/AggregatedExecutionStatistics.java
```diff
@@ -80,7 +79,8 @@ private static Model createAggregatedModel(List<HttpWorker.ExecutionStats> data,
         m.add(queryRes, IPROP.timeOuts, ResourceFactory.createTypedLiteral(timeOuts));
         m.add(queryRes, IPROP.wrongCodes, ResourceFactory.createTypedLiteral(wrongCodes));
         m.add(queryRes, IPROP.unknownException, ResourceFactory.createTypedLiteral(unknownExceptions));
-        m.add(queryRes, IPROP.totalTime, ResourceFactory.createTypedLiteral(toXSDDurationInSeconds(totalTime)));
+        m.add(queryRes, IPROP.totalTime, TimeUtils.createTypedDurationLiteralInSeconds(totalTime));
+        // m.add(queryRes, IPROP.totalTime, TimeUtils.createTypedDurationLiteral(totalTime));
```
bigerl, line 83: Add an TODO to commented out code so that we remember what it was all about.

bigerl (changes requested):
Looks good. Just some little things.

src/main/java/org/aksw/iguana/commons/time/TimeUtils.java
nck-mlcnv: No, it's not necessary. I just added in, because I thought that it would make it easier to compare mutliple durations by converting them to the same time unit. But I can remove it if you want

src/test/java/org/aksw/iguana/cc/query/list/QueryListTest.java
nck-mlcnv: The constructor of FileBasedQueryList doesn't throw NullPointerException on creation. It only does if you call an additional method. I just added that in quickly to make the test work, but I think explicitly checking for empty parameters in the constructors would be the better thing to do.

src/main/java/org/aksw/iguana/cc/metrics/impl/AggregatedExecutionStatistics.java
nck-mlcnv: I'm probably going to remove the toXSDDurationInSeconds method, because it returned an XSDDuration object which was then immediately converted into a typed literal. I think the method should just return the typed literal itself, like the `createTypedDurationLiteral` method did before.
When we're reverting the change, I will probably keep the `createTypeDurationLiteralInSeconds` method and update it. 

The toXSDDurationInSeconds method can be removed I think.

src/main/java/org/aksw/iguana/commons/time/DurationLiteral.java
```diff
@@ -0,0 +1,90 @@
...
+
+import java.time.Duration;
+
+/**
+ * This class is used to convert a Java Duration object to a typed RDF literal. The literal is typed as xsd:duration.<br/>
+ * TODO: This class temporarily fixes an issue with Jena.
+ */
+public class DurationLiteral implements RDFDatatype {
+
+    private final Duration duration;
```
mcb5637, line 16: Proper implementation of xsd::duration requires a value for days down to milliseconds and a second one for years and months. (because a conversion months<->days is not clearly defined)
(https://www.w3.org/TR/xmlschema11-2/#duration)
Although if you do not need months in your usecase, using xsd:dayTimeDuration (https://www.w3.org/TR/xmlschema11-2/#dayTimeDuration) might be more appropriate.

mcb5637 (changes requested):
As far as i can tell the spec of java.time.Duration matches xsd:dayTimeDuration and not xsd:duration.

src/main/java/org/aksw/iguana/commons/time/DurationLiteral.java
bigerl: dayTimeDuration sounds like a good fit for our use-case

nck-mlcnv: should we change it indefinitely to dayTimeDuration?

bigerl (approved):

PR #248, opened 2024-03-21
Add support for ahead-of-time compilation

nck-mlcnv:


bigerl (changes requested):
Thx for the PR. It looks good in general. Please find my questions and change requests below.

pom.xml
```diff
@@ -79,19 +74,9 @@
...
-        <dependency>
-            <groupId>org.apache.logging.log4j</groupId>
-            <artifactId>log4j-slf4j-impl</artifactId>
-            <version>${log4j.version}</version>
-        </dependency>
-        <dependency>
-            <groupId>org.apache.logging.log4j</groupId>
-            <artifactId>log4j-core</artifactId>
-            <version>${log4j.version}</version>
+            <groupId>ch.qos.logback</groupId>
```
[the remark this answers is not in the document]
nck-mlcnv: Log4j doesn't work with Graal

pom.xml
```diff
@@ -145,30 +130,13 @@
...
-        <dependency>
-            <groupId>org.springframework.data</groupId>
-            <artifactId>spring-data-commons</artifactId>
-            <version>3.1.2</version>
-        </dependency>
-        <dependency>
-            <groupId>org.springframework</groupId>
-            <artifactId>spring-context</artifactId>
-            <version>6.0.11</version>
-        </dependency>
```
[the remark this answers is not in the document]
nck-mlcnv: They were unused

graalvm/suite.yml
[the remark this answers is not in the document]
nck-mlcnv: For the native binary to work, you need to define some configurations. Instead of manually declaring these configurations I decided to rely on the tracing agent to automatically generate these configurations. 
The suite is configured to contain as many features as possible, so that the tracing agent creates the necessary configuration for all of them.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,156 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up JDK 17
+        uses: actions/setup-java@v3
+        with:
+          java-version: '17'
+          distribution: 'adopt'
+          cache: 'maven'
+      - name: Publish package
```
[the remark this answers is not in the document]
nck-mlcnv: I didn't quite understand the question. I just left this step as it was before.

src/main/java/org/aksw/iguana/cc/storage/impl/TriplestoreStorage.java
```diff
@@ -75,7 +79,11 @@ public void storeResult(Model data) {
 		//submit Block to Triple Store
 		UpdateProcessor processor = UpdateExecutionFactory
 				.createRemote(blockRequest, endpoint, createHttpClient());
-		processor.execute();
+		try {
+			processor.execute();
+		} catch (Exception e) {
+			logger.error("Error while storing data in triplestore: " + e.getMessage());
```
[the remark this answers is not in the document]
nck-mlcnv: The RDFFileStorage and CSVStorage are written in a similar way. Other than crashing, I wouldn't know what else should happen when the exception occurs. I added this, so that iguana doesn't just crash during the run, where graalvm generates the configurations.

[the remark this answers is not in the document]
bigerl: > I added this, so that iguana doesn't just crash during the run, where graalvm generates the configurations.

This refers to profiling for graalvm, right?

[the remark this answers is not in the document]
nck-mlcnv: Not for the profiling, but for when you execute the generate-config.sh script. The configuration generated by the tracing agent.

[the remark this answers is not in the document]
bigerl: But why should it crash?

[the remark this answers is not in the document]
nck-mlcnv: Because the script just tries to run every feature of iguana, but it doesn't actually benchmark a triplestore and there is also no triplestore where the data can be stored with the TriplestoreStorage. That's why it throws an exception at that point.

pom.xml
```diff
@@ -234,8 +234,57 @@
...
+                                    <executable>${project.basedir}/graalvm/generate-config.sh</executable>
+                                    <arguments>
+                                        <argument>-t</argument>
+                                        <argument>${project.build.directory}</argument>
+                                    </arguments>
+                                </configuration>
+                            </execution>
+                            <execution>
+                                <id>cleanup-files</id>
+                                <phase>test</phase>
```
bigerl, line 269: why do you run that in the test phase?

nck-mlcnv:
Symlink is fixed now.

graalvm/generate-config.sh
```diff
@@ -6,34 +6,48 @@ if [ -z "$GRAALVM_HOME" ]; then
...
+fi
+
+# Move generated configuration files from tests to the resources
+if [ -f "$TARGET_DIR"/native/agent-output/test/resource-config.json ]; then
+  mv "$TARGET_DIR"/native/agent-output/test/* src/main/resources/META-INF/native-image/
+fi
+
+"$GRAALVM_HOME"/bin/java -agentlib:native-image-agent=config-merge-dir=src/main/resources/META-INF/native-image/ -jar "$TARGET_DIR"/iguana.jar --help      > /dev/null
+"$GRAALVM_HOME"/bin/java -agentlib:native-image-agent=config-merge-dir=src/main/resources/META-INF/native-image/ -jar "$TARGET_DIR"/iguana.jar -is "$SUITE" > /dev/null
+"$GRAALVM_HOME"/bin/java -agentlib:native-image-agent=config-merge-dir=src/main/resources/META-INF/native-image/ -jar "$TARGET_DIR"/iguana.jar "$SUITE"     > /dev/null
```
bigerl, line 47: If I get this right, you are running all possible combinations of commandline arguments. Is that right? Please add a comment that references the section of the GraalVM docu that explains why this is necessary. (And add, if further helpful, an additional comment of your own)

bigerl (changes requested):
thx. changes look overall good. just two more little things.

pom.xml
nck-mlcnv: If you want to run the test as a native-binary, you can execute `mvn -Pnative -Dagent=true test`. This will first run the tests normally with the tracing agent from graalvm. After the test run, the native binary is compiled with the generated configurations from the previous run. 
This happens during the test phase of the native profile. But because of the bug, where the generated configurations of the tracing agent are wrongly formatted, this command needs to be executed first to cleanup the configurations, before they can be used for compilation.

graalvm/generate-config.sh
nck-mlcnv: I don't think the documentation explicitly says, that this is necessary. I just couldn't figure out how to configure this manually to be honest, so I just implemented it this way. 
Implementing it this way might also be better, because the configurations don't need to be manually updated, everytime something relevant in the code changes.

src/main/java/org/aksw/iguana/cc/storage/impl/TriplestoreStorage.java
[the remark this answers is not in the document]
bigerl: I think it would be better if it had an commandline option that is not documented in `--help` that enables this behavior for GraalVM instead of making it default behavior.

graalvm/generate-config.sh
bigerl: ok

src/main/java/org/aksw/iguana/cc/storage/impl/TriplestoreStorage.java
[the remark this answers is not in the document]
nck-mlcnv: I doesn't seen to be possible to hide the option from the  `--help` documentation. Should I add it anyway?

src/main/java/org/aksw/iguana/cc/worker/impl/SPARQLProtocolWorker.java
```diff
@@ -127,21 +129,30 @@ public SPARQLProtocolWorker(long workerId, ResponseBodyProcessor responseBodyPro
...
     public static void initHttpClient(int threadCount) {
         connectionManager = PoolingAsyncClientConnectionManagerBuilder.create()
-                .setMaxConnTotal(threadCount)
-                .setMaxConnPerRoute(threadCount)
+                .setMaxConnTotal(threadCount * 1000)
+                .setMaxConnPerRoute(threadCount * 1000)
+                .setPoolConcurrencyPolicy(PoolConcurrencyPolicy.LAX)
+                .setConnPoolPolicy(PoolReusePolicy.FIFO)
+                .setDefaultConnectionConfig(org.apache.hc.client5.http.config.ConnectionConfig.custom()
+                        .setConnectTimeout(Timeout.ofSeconds(5))
```
bigerl, line 137: Is it in any way important to set these values?

src/main/java/org/aksw/iguana/cc/worker/impl/SPARQLProtocolWorker.java
```diff
@@ -127,21 +129,30 @@ public SPARQLProtocolWorker(long workerId, ResponseBodyProcessor responseBodyPro
...
+                        .build())
                 .build();
         final var ioReactorConfig = IOReactorConfig.custom()
                 .setTcpNoDelay(true)
-                .setIoThreadCount(threadCount)
+                .setIoThreadCount(threadCount + 1)
```
bigerl, line 142: why +1? please document

src/main/java/org/aksw/iguana/cc/worker/impl/SPARQLProtocolWorker.java
```diff
@@ -127,21 +129,30 @@ public SPARQLProtocolWorker(long workerId, ResponseBodyProcessor responseBodyPro
...
+                .setSelectInterval(TimeValue.ofMilliseconds(100))
```
bigerl, line 143: what is this 0.1s interval doing?

src/main/java/org/aksw/iguana/cc/worker/impl/SPARQLProtocolWorker.java
```diff
@@ -127,21 +129,30 @@ public SPARQLProtocolWorker(long workerId, ResponseBodyProcessor responseBodyPro
...
+                .setSoKeepAlive(true)
                 .build();
         httpClient = HttpAsyncClients.custom()
                 .setConnectionManager(connectionManager)
                 .setIOReactorConfig(ioReactorConfig)
                 .setKeepAliveStrategy(new DefaultConnectionKeepAliveStrategy())
+                .setConnectionManagerShared(false)
```
bigerl, line 150: Is this important for something? Please document.

src/main/java/org/aksw/iguana/cc/worker/impl/SPARQLProtocolWorker.java
```diff
@@ -127,21 +129,30 @@ public SPARQLProtocolWorker(long workerId, ResponseBodyProcessor responseBodyPro
...
                 .setDefaultRequestConfig(RequestConfig.custom()
                         .setContentCompressionEnabled(false)
                         .setHardCancellationEnabled(true)
                         .build())
+                .setRetryStrategy(DefaultHttpRequestRetryStrategy.INSTANCE)
```
bigerl, line 155: This means one retry after 1s, see https://hc.apache.org/httpcomponents-client-5.2.x/current/httpclient5/apidocs/org/apache/hc/client5/http/impl/DefaultHttpRequestRetryStrategy.html#DefaultHttpRequestRetryStrategy--

I think there should be no retries. If the server fails to answer the query is failed.

src/main/java/org/aksw/iguana/cc/worker/impl/SPARQLProtocolWorker.java
```diff
@@ -345,11 +346,13 @@ protected int capacityIncrement() {
              */
             @Override
             protected void data(ByteBuffer src, boolean endOfStream) throws IOException {
-                if (endOfStream) {
+                if (endOfStream)
                     responseEnd = System.nanoTime();
-                    return;
```
bigerl, line 350: good catch

src/main/java/org/aksw/iguana/cc/worker/impl/SPARQLProtocolWorker.java
```diff
@@ -435,10 +445,22 @@ protected HttpExecutionResult buildResult() {
...
             // and it might be so small that it causes issues.
             return future.get(config.timeout().toNanos(), TimeUnit.NANOSECONDS);
-        } catch (InterruptedException | ExecutionException | TimeoutException e) {
+        } catch (InterruptedException | ExecutionException e) {
             // This will close the connection and cancel the request if it's still running.
             future.cancel(true);
             return createFailedResultBeforeRequest(queryIndex, e);
+        } catch (TimeoutException e) {
+            if (future.isDone()) {
+                LOGGER.warn("Request was already done after timeout.");
```
bigerl, line 454: I think this will not be helpful on the console. Bettwer would be:
```
Request finished immedeatly after timeout but will still be counted as timed out. 
```

src/test/java/org/aksw/iguana/cc/worker/impl/SPARQLProtocolWorkerTest.java
```diff
@@ -46,7 +46,16 @@ public class SPARQLProtocolWorkerTest {
...
+            .options(new WireMockConfiguration()
+                    .useChunkedTransferEncoding(Options.ChunkedEncodingPolicy.NEVER)
+                    .dynamicPort()
+                    .maxRequestJournalEntries(1000)
+                    .maxLoggedResponseSize(1000)
+                    .notifier(new ConsoleNotifier(false))
+                    .containerThreads(8)
+                    .asynchronousResponseEnabled(false)
+                    .asynchronousResponseThreads(8)
+                    .timeout(5000))
```
bigerl, line 58: What are the new options for? please document them.

src/test/java/org/aksw/iguana/cc/worker/impl/SPARQLProtocolWorkerTest.java
```diff
@@ -58,11 +69,12 @@ public class SPARQLProtocolWorkerTest {
     public static void setup() throws IOException {
         queryFile = Files.createTempFile("iguana-test-queries", ".tmp");
         Files.writeString(queryFile, QUERY, StandardCharsets.UTF_8);
-        SPARQLProtocolWorker.initHttpClient(1);
+        wm.setGlobalFixedDelay(5);
```
bigerl, line 72: What is the delay intended to trigger?

src/main/java/org/aksw/iguana/cc/worker/impl/SPARQLProtocolWorker.java
nck-mlcnv: The MaxConnTotal and MaxConnPerRoute are important. The rest were kind of left over. They may increase the performance, but I'm not sure.

nck-mlcnv: The values for MaxConnTotal and MaxConnPerRoute are kind of arbitrary though. They need to be set to a high value, but I don't know how they correlate to the problems that occured.

src/test/java/org/aksw/iguana/cc/worker/impl/SPARQLProtocolWorkerTest.java
nck-mlcnv: One of the tests were failing because of wiremock. It seemed to lock itself due to high workload. (i think)
Limiting the rate at which request are processed by adding a fixed delay seemed to fix the problem.
The original problem only occured on the github runners.

src/main/java/org/aksw/iguana/cc/controller/MainController.java
```diff
@@ -28,14 +28,16 @@ public Path convert(String value) {
         @Parameter(names = {"--ignore-schema", "-is"}, description = "Do not check the schema before parsing the suite file.")
         private boolean ignoreShema = false;
 
+        @Parameter(names = {"--dry-run", "-d"}, hidden = true)
+        public static boolean dryRun = false;
+
```
bigerl, line 33: If I remember correctly you said that hidden didn't work. 
Thus, as suggested, please look in argc for `--dry-run` and `-d` instead to set the option.

src/main/java/org/aksw/iguana/cc/storage/impl/TriplestoreStorage.java
```diff
@@ -79,10 +81,14 @@ public void storeResult(Model data) {
 		//submit Block to Triple Store
 		UpdateProcessor processor = UpdateExecutionFactory
 				.createRemote(blockRequest, endpoint, createHttpClient());
-		try {
+		if (MainController.Args.dryRun) {
```
bigerl, line 84: Please add a comment that explains this branching. Otherwise, it will be hard to recall what it was good for in the future.

src/main/java/org/aksw/iguana/commons/io/ByteArrayListInputStream.java
```diff
@@ -0,0 +1,158 @@
...
+        this.iterator = data.iterator();
+        if (iterator.hasNext()) {
+            this.currentBuffer = ByteBuffer.wrap(iterator.next());
+        } else {
+            this.currentBuffer = null;
+        }
+    }
+
+    private void updateAvailable(long n) {
+        if (available != null) {
```
bigerl, line 31: should this have a check for n>available?

bigerl (changes requested):
Generally looks good. Nice work with the ByteArrayList streams. Only some minor things left, see below.

src/main/java/org/aksw/iguana/cc/controller/MainController.java
nck-mlcnv: Sorry, for the confusion. What I meant was, that I don't think it was possible to hide an option with JCommander, but I seem to have overlooked the `hidden`attribute that I used there, which does exactly what was needed. :)

src/main/java/org/aksw/iguana/commons/io/ByteArrayListInputStream.java
nck-mlcnv: I think removing the variable and the method would be better. The available method isn't called that often, so recalculating the available bytes shouldn't be bad.

bigerl: also fine with me.

src/main/java/org/aksw/iguana/cc/storage/impl/TriplestoreStorage.java
```diff
@@ -81,6 +81,11 @@ public void storeResult(Model data) {
 		//submit Block to Triple Store
 		UpdateProcessor processor = UpdateExecutionFactory
 				.createRemote(blockRequest, endpoint, createHttpClient());
+
+		// If dry run is enabled, the data will not be send to an existing triplestore,
```
bigerl, line 85:
```suggestion
		// If dry run is enabled, the data will not be sent to an existing triplestore,
```

bigerl (approved):

PR #251, opened 2024-06-02
Add missing dataset property to example

nck-mlcnv:


bigerl:
The changes look fine. But why should we make the dataset configuration optional? Without providing information about the dataset the result files will be less self-contained.

nck-mlcnv:
I left out the dataset in one of the exmples. I'm reusing this branch to add that in.

bigerl (approved):

PR #254, opened 2024-06-13
Update result data format

nck-mlcnv:


bigerl (approved):
Looks good 👍🏻

src/main/java/org/aksw/iguana/cc/tasks/impl/StresstestResultProcessor.java
```diff
@@ -143,6 +143,14 @@ public void calculateAndSaveMetrics(Calendar start, Calendar end) {
             }
         }
 
+        // Create Query nodes with their respective queryIDs
+        for (String queryID : queryIDs) {
+            Resource queryRes = IRES.getResource(queryID);
+            m.add(queryRes, RDF.type, IONT.query);
+            m.add(queryRes, IPROP.fullID, ResourceFactory.createTypedLiteral(queryID));
+            m.add(queryRes, IPROP.id, ResourceFactory.createTypedLiteral(BigInteger.valueOf(Long.parseLong(queryID.split(":")[1]))));
```
bigerl, line 151: This looks dangerous. How does queryID look like and how is guaranteed that it contains only one colon?

It is a IRI, right? It might have a protocol and a port, both using a colon: `http://example.com:8000/`

bigerl (changes requested):
Otherwise, the changes look good. Please use protege to verify the ontology and to validate an output file with the ontology.

src/main/java/org/aksw/iguana/cc/tasks/impl/StresstestResultProcessor.java
nck-mlcnv: QueryID consists of the hashcode of its query handler (the list of queries it's a part of to be precise) and the id inside of that list in this format: <queryhandler_hash>:<id>

schema/iguana.owl
```diff
@@ -162,42 +195,152 @@
...
     <!-- DatatypeProperty definitions -->
 
+  <owl:DatatypeProperty rdf:about="&iprop;time">
+    <rdfs:label>time</rdfs:label>
+    <rdfs:comment xml:lang="en">The time duration of the execution of a query.</rdfs:comment>
+    <rdfs:domain rdf:resource="&iont;QueryExecution"/>
+    <rdfs:range rdf:resource="http://www.w3.org/2001/XMLSchema#dayTimeDuration"/>
+  </owl:DatatypeProperty>
+
+  <owl:DatatypeProperty rdf:about="&iprop;startTime">
```
bigerl, line 260: Would be better to drop this and use startDate here as well instead.

schema/iguana.owl
```diff
@@ -162,42 +195,152 @@
...
+        </owl:unionOf>
+      </owl:Class>
+    </rdfs:domain>
     <rdfs:range rdf:resource="&iont;Query"/>
-  </owl:ObjectProperty>  
+  </owl:ObjectProperty>
 
     <!-- DatatypeProperty definitions -->
 
+  <owl:DatatypeProperty rdf:about="&iprop;time">
```
bigerl, line 253: Please check for all properties if they have domain and range set. (only for vocabulary defined here not for imported)

bigerl, line 253: executionTook would be better property names.

schema/iguana.owl
```diff
@@ -1,4 +1,3 @@
-
 <!DOCTYPE rdf:RDF [
     <!ENTITY dcterms "http://purl.org/dc/terms/" >
     <!ENTITY owl "http://www.w3.org/2002/07/owl#" >
```
bigerl, line 3: Please add the following imports:
```
        <owl:imports rdf:resource="http://purl.org/dc/elements/1.1/"/>
        <owl:imports rdf:resource="http://purl.org/dc/terms/"/>
        <owl:imports rdf:resource="http://www.w3.org/1999/02/22-rdf-syntax-ns#"/>
        <owl:imports rdf:resource="http://www.w3.org/2000/01/rdf-schema#"/>
        <owl:imports rdf:resource="http://www.w3.org/2002/07/owl"/>
```

in `<owl:Ontology ...>`

Also, please double check if I missed any. Ontologies do typically not import any vocabulary.

bigerl, line 3: Please double check what a good ontology name would be. This one with the owl file seems not ideal to me. 

Something like `https://iguana-benchmark.eu/ontology/` or even https://vocab.dice-research.org/iguana/` seems to fit better. It would also make sense to have a version iri like `https://vocab.dice-research.org/iguana/4.0.0/`

Also, `iont`, `iprop` and `ires` should be relative to that.

If we define the ontology to live under dice it should be published there. That is a follow up task. Please contact @lennartaustenfeld about that.

schema/iguana.owl
```diff
@@ -1,4 +1,3 @@
-
```
bigerl, line 1: I think in general it would make sense to use the Protégé generated file hear instead of the manually written one. This is RDF/XML if I am not mistaken (so not OWL/XML).

bigerl (changes requested):
Another batch of change requests

schema/iguana.owl
nck-mlcnv: I change the name to `https://iguana-benchmark.eu/ontology/` for now and added the version iri `https://vocab.dice-research.org/iguana/4.0.0/`.

I made `iont`, `iprop` and `ires` relative to `https://iguana-benchmark.eu/ontology/`, i don't know if that's correct.

nck-mlcnv: I've also created an issue for possible renaming and for making the ontology available online.

schema/iguana.owl
nck-mlcnv: They should all have a domain and range. Protege doesn't show the domain and range of a property, if it has xsd:dayTimeDuration as its range for some reason.

nck-mlcnv: Change from xsd:dayTimeDuration to xsd:duration. Protege still doesn't show them, but they should be there.

schema/iguana.owx
```diff
@@ -0,0 +1,1395 @@
...
+    <Prefix name="rdf" IRI="http://www.w3.org/1999/02/22-rdf-syntax-ns#"/>
+    <Prefix name="xml" IRI="http://www.w3.org/XML/1998/namespace"/>
+    <Prefix name="xsd" IRI="http://www.w3.org/2001/XMLSchema#"/>
+    <Prefix name="iont" IRI="https://iguana-benchmark.eu/ontology/class/"/>
+    <Prefix name="ires" IRI="https://iguana-benchmark.eu/ontology/resource/"/>
+    <Prefix name="rdfs" IRI="http://www.w3.org/2000/01/rdf-schema#"/>
+    <Prefix name="skos" IRI="http://www.w3.org/2004/02/skos/core#"/>
+    <Prefix name="iprop" IRI="https://iguana-benchmark.eu/ontology/properties/"/>
+    <Prefix name="schor" IRI="https://w3id.org/scholarlydata/role/"/>
+    <Prefix name="dcterms" IRI="http://purl.org/dc/terms/"/>
```
bigerl, line 21: Did you add the prefixes or was that Protege?

Some of them are never used, i.e. schor and skos

bigerl (approved):
Looks good. There is just this one oddity with the prefixes. But nothing serious.

schema/iguana.owx
nck-mlcnv: They were already added before I changed the ontology so I just kept them. I don't know who/what added them.
Should I remove them?

nck-mlcnv: I removed them.

nck-mlcnv:
The failing test can probably be ignored, should be fixed already in another branch👍

PR #256, opened 2024-07-24
Update Documentation Deployment

nck-mlcnv:
Mostly just a cleanup of the old deployment:
- Remove unnecessary images and files
- Upadte mkdocs.yaml file
- Add deployment of latest documentations
- Small changes to documentation


.github/workflows/deploy.yml
bigerl, line 1: I would like to have something like a test deployment where we can double check the result of a change before we merge into develop. (non-blocking)

bigerl (approved):
From what I can see this looks good. Please make sure that it works, so, e.g., temporarily change the target branch before you merge. I would like to avoid debugging on main.

mkdocs.yml
```diff
@@ -62,33 +58,17 @@ color: "#82BDC3"
...
 nav:
   - "General":
-    - "About": "about.md"
-    - "Architecture": "architecture.md"
-
-  - "Quick Start Guide":
-    - "Download": "download.md"
-    - "Quick Configuration": "quick-config.md"
-    - "Run Iguana": "run-iguana.md"
+    - "About": "README.md"
```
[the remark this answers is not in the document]
nck-mlcnv: Yes, should be possible, but also difficult, because the file probably doesn't change that often and you'd have to check for each deployment if the file has changed or not.

.github/workflows/deploy.yml
nck-mlcnv: I added a test worfklow that just uploads the artifacts instead of deploying them. I don't know if that was the correct thing you meant.

nck-mlcnv: The compile native test should already be fixed in the develop branch.

mkdocs.yml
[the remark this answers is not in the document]
nck-mlcnv: nevermind, added it now

.github/workflows/deploy.yml
bigerl, line 1: There are still some differences between this and https://github.com/dice-group/IGUANA/blob/422e7d6b7ed4920bbc52cf497f2475a5fe520eab/.github/workflows/deploy-test.yml

Please use IntelliJ's right-click->compare with function on the commit before you deleted deploy-test (422e7d6b7ed4920bbc52cf497f2475a5fe520eab) to resolve these.

bigerl (changes requested):

bigerl (approved):

PR #257, opened 2024-07-25
Fix github workflow

nck-mlcnv:


bigerl (changes requested):
Changes look good but please figure out why the tests workflow is skipped. The PR goes into develop, so it should be executed, right?

bigerl:
I would rather like to see it run before we merge. Tests from previous commits on the PR should be canceled. I've sent you a PM with an example.

bigerl (approved):

PR #258, opened 2024-07-26
Update GitHub pages

nck-mlcnv:
* Remove all unsued files
* Add latest path for documentation
* Website only shows documentation now


bigerl:
I am a bit confused. Isn't this something that should automatically be generated?
Also, I am not completely sure what I should review here.

nck-mlcnv:
Only the documentation will be automatically generated. The pages currently includes more than the documentations, the homepage for example, and other things that aren't used at all (https://iguana-benchmark.eu/sc/). 
The home page is also quite outdated and hard to maintain (there are also no source files available).

That's why I think, that the website should be exclusively used for the documentation for now. The readme of the github repository should also be enough as an replacement for a website in my opinion. It doesn't really add anything extra anyway.

PR #265, opened 2024-08-07
Upload fixed documentation page

nck-mlcnv:


docs/4.0.0/404.html
```diff
@@ -12,15 +12,15 @@
       
       
       <link rel="icon" href="/IGUANA_logo.png">
-      <meta name="generator" content="mkdocs-1.6.0, mkdocs-material-9.5.31">
+      <meta name="generator" content="mkdocs-1.5.3, mkdocs-material-9.5.15">
```
bigerl, line 15: this looks wrong

bigerl (changes requested):
I need some more info about this change as I will obviously not review it line-by-line. 

Did you generate it locally?

docs/4.0.0/404.html
nck-mlcnv: Yeah, I generated it locally, I used the wrong version

bigerl (approved):

PR #266, opened 2024-08-07
Fix issues with queries that are bigger than 2GB

nck-mlcnv:


nck-mlcnv:
There are changes done to the BigByteArrayInputStream, that might not be necessary anymore. Should I revert them?

docs/configuration/big_queries.md
bigerl, line 1: rename file to huge_query_strings.md

docs/configuration/big_queries.md
```diff
@@ -0,0 +1,5 @@
+# Reading Large Queries
+
+When working with large queries (Queries that are larger than 2^31 Bytes or ~2GB),
```
bigerl, line 3:
```suggestion
# Huge Query Strings

When working with large query strings (longer than 2³² Bytes or ~2GB), 
```

src/main/java/org/aksw/iguana/commons/io/ByteArrayListInputStream.java
```diff
@@ -125,12 +141,16 @@ public void skipNBytes(long n) throws IOException {
     }
 
     @Override
-    public int available() throws IOException {
+    public int available() {
         return (int) Math.min(Integer.MAX_VALUE, availableLong());
```
bigerl, line 145:
```suggestion
        return (int) Math.min(Integer.MAX_VALUE - 8, availableLong());
```

Would probably be good to introduce a constant `MAX_ARRAY_SIZE = Integer.MAX_VALUE - 8` for that

src/main/java/org/aksw/iguana/commons/io/ByteArrayListInputStream.java
```diff
@@ -95,6 +99,18 @@ public int readNBytes(byte[] b, int off, int len) throws IOException {
         return read(b, off, len, 0);
```
bigerl, line 99: Throw IOExceltion if `length>MAX_ARRAY_SIZE. Also in read(byte[] b, int off, int len)`.
From how I understand https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/io/InputStream.html the length must only be truncated if the end is reached.

src/main/java/org/aksw/iguana/commons/io/BigByteArrayInputStream.java
```diff
@@ -51,71 +57,81 @@ public int read(byte[] b, int off, int len) throws IOException {
...
+        if (currentBuffer == null && !activateNextBuffer()) return new byte[0];
+        int read = Math.min(n, available());
+        byte[] b = new byte[read];
+        readNBytes(b, 0, read);
+        return b;
+    }
+
     @Override
     public int readNBytes(byte[] b, int off, int len) throws IOException {
         this.bbaos.close();
```
bigerl, line 91: check len against MAX_ARRAY_SIZE

src/main/java/org/aksw/iguana/commons/io/BigByteArrayInputStream.java
```diff
@@ -51,71 +57,81 @@ public int read(byte[] b, int off, int len) throws IOException {
...
             activateNextBuffer();
 
         return copyLength1 + copyLength2;
     }
 
     @Override
     public byte[] readAllBytes() throws IOException {
-        throw new IOException("Reading all bytes from a BigByteArrayInputStream is prohibited because it might exceed the array capacity");
+        if (currentBuffer == null && !activateNextBuffer()) return new byte[0];
+        if (availableLong() > Integer.MAX_VALUE - 8)
```
bigerl, line 118: use constant

src/main/java/org/aksw/iguana/commons/io/BigByteArrayInputStream.java
```diff
@@ -131,33 +147,35 @@ private boolean activateNextBuffer() {
...
-            return ended = activateNextBuffer();
+        if (currentBuffer.remaining() == 0)
+            return ended = !activateNextBuffer();
         ended = false;
         return true;
     }
 
+    @Override
+    public int available() {
+        return (int) Math.min(availableLong(), Integer.MAX_VALUE);
```
bigerl, line 170: MAX_ARRAY_SIZE

bigerl (changes requested):
Thx for the PR. Looks good to me. I just commented some minor issues below. 
Please also provide benchmarks comparing this with develop as a baseline to make sure we don't introduce a performance regression here. Please use the GraalVM binaries.

src/main/java/org/aksw/iguana/commons/io/BigByteArrayInputStream.java
nck-mlcnv: Is this necessary? Because if the given array is larger than MAX_ARRAY_SIZE, iz might not be able to exist in the first place because it's large than MAX_ARRAY_SIZE. Otherwise if the len and array size don't fit, the method will throw anyway because of `Objects.checkFromIndexSize(off, len, b.length);`.

docs/configuration/big_queries.md
nck-mlcnv: I moved the section to the file `queries.md`.

bigerl (approved):

PR #267, opened 2024-08-16
Fix the response body processor timeout message

nck-mlcnv:
The timeout message for the response body processor hash a bad formatting for the time.

### DIFF
--- a/src/main/java/org/aksw/iguana/cc/worker/ResponseBodyProcessor.java
+++ b/src/main/java/org/aksw/iguana/cc/worker/ResponseBodyProcessor.java
@@ -72,7 +72,10 @@ public class ResponseBodyProcessor {
             return responseDataMetrics;
         }
 
-        LOGGER.info(MessageFormat.format("Shutting down ResponseBodyProcessor with {0} min timeout to finish processing. {1} tasks remaining.", timeout.toMinutes() + "." + (timeout.toSecondsPart() / 6), executor.getQueue().size()));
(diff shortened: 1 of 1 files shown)


src/main/java/org/aksw/iguana/cc/worker/ResponseBodyProcessor.java
```diff
@@ -72,7 +72,7 @@ public List<LanguageProcessor.LanguageProcessingData> getResponseDataMetrics() {
             return responseDataMetrics;
         }
 
-        LOGGER.info(MessageFormat.format("Shutting down ResponseBodyProcessor with {0} min timeout to finish processing. {1} tasks remaining.", timeout.toMinutes() + "." + (timeout.toSecondsPart() / (double) 60), executor.getQueue().size()));
+        LOGGER.info(MessageFormat.format("Shutting down ResponseBodyProcessor with {0} min timeout to finish processing. {1} tasks remaining.", timeout.toMinutes() + "." + (timeout.toSecondsPart() / 6), executor.getQueue().size()));
```
bigerl, line 75: Why divide by 6? This seems more like it should be:

```suggestion
        LOGGER.info(MessageFormat.format("Shutting down ResponseBodyProcessor with {0}min {1}s timeout to finish processing. {2} tasks remaining.",
                timeout.toMinutes(), 
                String.format("%02d", timeout.toSecondsPart() % 60),
                executor.getQueue().size()));
                ```

bigerl (changes requested):

nck-mlcnv:
`toSecondsPart()` already delivers the seconds part with a value between 0 and 59.

bigerl (approved):

PR #269, opened 2024-08-19
Change Ontology IRI

nck-mlcnv:
Also does:
- removes unused test files
- add file upload of the ontology to the ftp server


bigerl (changes requested):
Thx for the PR. This seems not fully completed yet. 

1. Is the distribution at that path clarified? A action that publishes the file on release would be good. 
2.Three are still quite some occurrences of `iguana-benchmark.eu` left:
image

nck-mlcnv:
The test files are all not used actually, I will remove them soon

bigerl:
I think the files should then be removed in this PR so that we don't have to change them.

.github/workflows/deployment_test.yml
```diff
@@ -51,7 +51,7 @@ jobs:
         uses: SamKirkland/FTP-Deploy-Action@v4.3.5
         with:
           server: files.dice-research.org
```
bigerl, line 53: it should be ftp over tls. Not sure how to set that up here

nck-mlcnv: Yeah I thought so too, but something didn't work on the first run

bigerl: maybe `ftps-legacy`?

nck-mlcnv: didn't work either

nck-mlcnv: I don't know if this is relevant but it seems to be a common [issue](https://github.com/SamKirkland/FTP-Deploy-Action/issues/106) . I think I will just try to use lftp itself.

bigerl (approved):
Changes look now good to me.

PR #273, opened 2024-09-05
Add query patterns

nck-mlcnv:


src/main/java/org/aksw/iguana/cc/query/handler/QueryHandler.java
```diff
@@ -224,4 +284,67 @@ public String[] getAllQueryIds() {
...
+            while (matcher.find()) {
+                final var match = matcher.group();
+                final var variable = "?" + match.replaceAll("%%", "");
+                variables.add(variable);
+                patternQueryString = patternQueryString.replaceAll(match, variable);
+            }
+
+            // build SELECT query for finding bindings for the variables
+            final var patternQuery = QueryFactory.create(patternQueryString);
+            final var whereClause = "WHERE " + patternQuery.getQueryPattern();
```
bigerl, line 324: This approach does not consider subqueries. 
Please state in the documentation that placeholders in subqueries must be projected from the subquery, otherwise they will not be processed correctly.  

If we really want to support subqueries that needs a bit more thought. We would probably need to extract them from the jena query object and send them separately. Also, we would need to handle deeply nested subqueries somehow. Seems like a deep rabi thole to me that I would like to avoid. 

@aalexandrasilva: Do you have placeholders in subqueries in your workload?

src/main/java/org/aksw/iguana/cc/query/handler/QueryHandler.java
```diff
@@ -224,4 +284,67 @@ public String[] getAllQueryIds() {
...
+        for (int i = 0; i < patternQueries.size(); i++) {
+            // replace all variables in the query pattern with SPARQL variables
+            // and store the variable names
+            var patternQueryString = patternQueries.getQuery(i);
+            final Matcher matcher = pattern.matcher(patternQueryString);
+            final var variables = new ArrayList<String>();
+            while (matcher.find()) {
+                final var match = matcher.group();
+                final var variable = "?" + match.replaceAll("%%", "");
+                variables.add(variable);
```
bigerl, line 318: The variable names might already been taken at this point. I am not sure if this is supposed to be a feature. I hope not. 

I would suggest to use `QueryFactory.create(patternQueryString)` to get the existing variable names and make sure to generate new ones. 


That also means that variables should be a map<String,String> from template name to variable name.

src/main/java/org/aksw/iguana/cc/query/handler/QueryHandler.java
```diff
@@ -224,4 +284,67 @@ public String[] getAllQueryIds() {
...
+                    LOGGER.warn("No results for query pattern: {}", patternQueryString);
+                }
+                while (resultSet.hasNext()) {
+                    var instance = new ParameterizedSparqlString(patternQueryString);
+                    QuerySolution solution = resultSet.next();
+                    for (String var : resultSet.getResultVars()) {
+                        instance.clearParam(var);
+                        instance.setParam(var, solution.get(var));
+                    }
+                    instances.add(instance.toString());
```
bigerl, line 344: We don't need more than limit instances. So, you can break the while look when that many were found.

src/main/java/org/aksw/iguana/cc/query/handler/QueryHandler.java
```diff
@@ -121,13 +136,21 @@ public String value() {
                 return value;
             }
         }
+
+        public record Pattern(@JsonProperty(required = true) URI endpoint, Long limit, Boolean save) {
+            public Pattern(URI endpoint, Long limit, Boolean save) {
+                this.endpoint = endpoint;
+                this.limit = limit == null ? 2000 : limit;
```
bigerl, line 143: Please document in the MD

src/main/java/org/aksw/iguana/cc/query/handler/QueryHandler.java
```diff
@@ -224,4 +284,67 @@ public String[] getAllQueryIds() {
...
+    * and will request query solutions from the given sparql endpoint (e.g DBpedia).<br/>
+    * The solutions will then be instantiated into the query pattern.
+    * The result may look like the following:<br/>
+    * <code>SELECT * WHERE {?s &lt;http://prop/1&gt; ?o . ?o &lt;http://exa.com&gt; "123"}</code><br/>
+    * <code>SELECT * WHERE {?s &lt;http://prop/1&gt; ?o . ?o &lt;http://exa.com&gt; "12"}</code><br/>
+    * <code>SELECT * WHERE {?s &lt;http://prop/2&gt; ?o . ?o &lt;http://exa.com&gt; "1234"}</code><br/>
+    */
+    private static List<String> instantiatePatternQueries(QuerySource querySource, Config.Pattern config) throws IOException {
+        final var patternQueries = new FileCachingQueryList(querySource);
+        final Pattern pattern = Pattern.compile("%%var\\d+%%");
```
bigerl, line 307: use `%%var[0-9]+%%` because it is easier to read and devs don't get confused when they find here a different regex than in the documentation.

src/test/java/org/aksw/iguana/cc/query/handler/PatternQueriesTest.java
```diff
@@ -0,0 +1,85 @@
...
+                    .dynamicPort()
+                    .notifier(new ConsoleNotifier(true))
+                    .jettyIdleTimeout(2000L)
+                    .jettyStopTimeout(2000L)
+                    .timeout(2000))
+            .build();
+
+    @Test
+    public void testPatternQueries() throws IOException {
+        String patternQuery = "PREFIX rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> SELECT * WHERE {?s rdf:type %%var0%% ; %%var1%% %%var2%%. %%var2%% ?p <http://www.w3.org/2002/07/owl#Thing>}";
```
bigerl, line 57: Please add a test for a malformed pattern, e.g.,
```
SELECT * WHERE {%%var0%% %%var0%% %%var0%% %%var0%%}
```

bigerl (changes requested):
Thx for the PR. General design seems good. I am also happy to see that apperently only very few/small changes to other parts were necessary to integrate this. So, nice work with the original v4 release. 👍 Please find my suggestions for imporvements below.

src/main/java/org/aksw/iguana/cc/query/handler/QueryHandler.java
aalexandrasilva: No, I don't have placeholders in the subqueries :smile:

schema/iguana-schema.json
```diff
@@ -335,9 +335,29 @@
...
+    "Pattern": {
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "endpoint": {
+          "type": "string"
+        },
+        "limit": {
+          "type": "integer",
+          "minimum": 1
```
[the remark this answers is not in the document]
nck-mlcnv: We don't need a default value here, it is just an assertion for the input values (no negative numbers).

src/main/java/org/aksw/iguana/cc/query/handler/QueryHandler.java
nck-mlcnv: I didn't quite understand what you mean by that, but QueryFactory.create(patternQueryString) probably wouldn't work because of the placeholders. I think using a Set instead of a List for storing the variables names should suffice.

docs/configuration/queries.md
```diff
@@ -93,3 +93,41 @@ tasks:
...
+  format: "folder" 
+  pattern:
+    endpoint: "http://dbpedia.org/sparql"
+    limit: 100
+    save: true
+```  
+
+Pattern queries are queries that contain placeholders.
+A query pattern is a SPARQL 1.1 Query, which can have additional variables in the regex form of
+`%%var[0-9]+%%` in the Basic Graph Pattern.
```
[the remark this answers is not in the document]
nck-mlcnv: I just commented on what `[0-9]+` represents. I don't know if that's enough.

src/main/java/org/aksw/iguana/cc/query/handler/QueryHandler.java
```diff
@@ -150,18 +173,55 @@ protected QueryHandler() {
...
-            case ONE_PER_LINE -> new FileLineQuerySource(Path.of(config.path()));
-            case SEPARATOR -> new FileSeparatorQuerySource(Path.of(config.path()), config.separator);
-            case FOLDER -> new FolderQuerySource(Path.of(config.path()));
-        };
+        this.config = config;
+        var querySource = createQuerySource(Path.of(config.path));
+
+        if (config.pattern() != null) {
+            final var originalPath = querySource.getPath();
+            Path instancePath = Files.isDirectory(originalPath) ?
```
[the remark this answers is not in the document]
nck-mlcnv: I just moved some code to a different method, maybe that helps.

docs/configuration/queries.md
```diff
@@ -94,40 +95,50 @@ tasks:
...
-```yaml
-queries:
-  path: "./example/suite/queries/"
-  format: "folder" 
-  pattern:
-    endpoint: "http://dbpedia.org/sparql"
-    limit: 100
-    save: true
-```  
+The placeholders are written in the form of `%%var[0-9]+%%`, where `[0-9]+` represents any number.
```
bigerl, line 103: As discussed on Tuesday that should rather be something like `%%[a-zA-Z0-9_]+%%`. 

Also mention, that even if the placeholder is equal to a variable name it will not be assigned the same variable name during candidate generation.

src/main/java/org/aksw/iguana/cc/query/handler/QueryHandler.java
```diff
@@ -176,37 +175,48 @@ public QueryHandler(Config config) throws IOException {
...
+
+    private QueryList initializeTemplateQueryHandler(QuerySource templateSource) throws IOException {
+        QuerySource querySource = templateSource;
+        final var originalPath = templateSource.getPath();
+        final Path instancePath = Files.isDirectory(originalPath) ?
+                originalPath.resolveSibling(originalPath.getFileName() + "_instances.txt") : // if the source of the query templates is a folder, the instances will be saved in a file with the same name as the folder
+                originalPath.resolveSibling(originalPath.getFileName().toString().split("\\.")[0] + "_instances.txt"); // if the source of the query templates is a file, the instances will be saved in a file with the same name as the file
+        if (Files.exists(instancePath)) {
+            LOGGER.info("Already existing query template instances have been found and will be reused. Delete the following file to regenerate them: {}", instancePath.toAbsolutePath());
+            querySource = createQuerySource(instancePath); // if the instances already exist, use them
```
bigerl, line 197: The path should contain not only the file-name but also contain the parameters:
endpoint and limit. For endpoint, use  a hash of the endpoint string. 

Can be moved to a separate low priority issue and resolved later.

src/main/java/org/aksw/iguana/cc/query/handler/QueryHandler.java
```diff
@@ -287,55 +297,73 @@ public Config getConfig() {
...
-    private static List<String> instantiatePatternQueries(QuerySource querySource, Config.Pattern config) throws IOException {
-        final var patternQueries = new FileCachingQueryList(querySource);
-        final Pattern pattern = Pattern.compile("%%var\\d+%%");
+    private static List<String> instantiateTemplateQueries(QuerySource querySource, Config.Template config) throws IOException {
+        // charset for generating random variable names
+        final String charset = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789";
+        final Random random = new Random();
+
+        final var templateQueries = new FileCachingQueryList(querySource);
+        final Pattern template = Pattern.compile("%%var[0-9]+%%");
```
bigerl, line 323:
```suggestion
        final Pattern template = Pattern.compile("%%[a-zA-Z0-9_]+%%");
```

src/main/java/org/aksw/iguana/cc/query/handler/QueryHandler.java
```diff
@@ -287,55 +297,73 @@ public Config getConfig() {
...
+            final Matcher matcher = template.matcher(templateQueryString);
+            final var variables = new LinkedHashMap<String, String>(); // a set, that preserves insertion order
             while (matcher.find()) {
                 final var match = matcher.group();
-                final var variable = "?" + match.replaceAll("%%", "");
-                variables.add(variable);
-                patternQueryString = patternQueryString.replaceAll(match, variable);
+                if (variables.containsKey(match)) continue;
+                String variableName = match.replaceAll("%%", "");
+                while (templateQueryString.contains("?" + variableName)) { // generate random variable name with 20 characters until it is unique
```
bigerl, line 335:
```suggestion
                while (templateQueryString.contains("?" + variableName) || templateQueryString.contains("$" + variableName)) { // generate random variable name with 20 characters until it is unique
```

$x and ?x are equivalent ways to write a variable. Don't ask me why they did this.

src/main/java/org/aksw/iguana/cc/query/handler/QueryHandler.java
bigerl: was clarified in meeting and is resolved now.

bigerl (changes requested):
Thx for the update. Almost there. This change request should be the last.

bigerl (approved):

PR #281, opened 2024-09-23
Send Update Queries to seperate endpoint

nck-mlcnv:


src/test/java/org/aksw/iguana/cc/query/QueryDataTest.java
```diff
@@ -0,0 +1,65 @@
...
+
+    @BeforeAll
+    public static void setup() throws IOException {
+        tempFile = Files.createTempFile("test", "txt");
+        Files.writeString(tempFile, """
+                SELECT ?s ?p ?o WHERE {
+                    ?s ?p ?o
+                }
+                
+                INSERT DATA {
```
bigerl, line 30: Just realized we are also missing something here. Please create an issue to also support multiple semicolon-chained update operations. 

https://www.w3.org/TR/2013/REC-sparql11-update-20130321/#updateLanguage

bigerl (approved):
Thx for the PR. This looks already quite polished. Please just create the issue metioned below.

src/test/java/org/aksw/iguana/cc/query/QueryDataTest.java
nck-mlcnv: What exactly is the problem with semicolon-chained update operations? Should they be split up into multiple queries?

bigerl: Good question. They should just work, right? Only, We should probably double check that LSQ handles them correctly when we include that.

bigerl (approved):

PR #283, opened 2024-09-26
Add result aggregation for query templates

nck-mlcnv:
The new default behaviour is, that each result from the instances of a query template will be counted as the result of the query template. This is done by forcing the query instances to use the id of the query template in the results.
The query templates are considered as queries as well, but they won't be executed.

The ids of those queries that won't be presented in the results (instance queries if `individualResults` is turned on, query templates otherwise) will be shifted to the right, outside of the range of the number of queries whose results will be presented.

The serialization of the query instaces needed to be changed as well, so that the query instances can be mapped back to their
templates.

The query handler also needs to differentiate between the number of executable queries and the number of queries, that
will be presented in the results, otherwise, too few queries might be executed.

Todo:
- [x] Tests
- [x] Documentation


nck-mlcnv:
Also, the result model probably needs to be changed, so that it includes the relation between query templates and their instances.

graalvm/suite.yml
```diff
@@ -6,7 +6,7 @@ connections:
   - name: "Blazegraph"
     version: "1.1.1"
     dataset: "DatasetName"
-    endpoint: "http://localhost:9999/blazegraph/sparql"
+    endpoint: "https://dbpedia.org/sparql"
```
bigerl: I think it would be better to use the [wikidata endpoint](https://query.wikidata.org/). I see it as more long-living. (same below)

src/main/java/org/aksw/iguana/cc/query/handler/QueryHandler.java
```diff
@@ -141,18 +144,18 @@ public String value() {
...
                 this.save = save == null || save;
+                this.individualResults = individualResults != null && individualResults;
             }
         }
     }
 
-    public record QueryStringWrapper(int index, String query, boolean update) {}
-
-    public record QueryStreamWrapper(int index, boolean cached, Supplier<InputStream> queryInputStreamSupplier, boolean update) {}
+    public record QueryStringWrapper(int index, String query, boolean update, Integer resultId) {}
```
bigerl, line 156: document what the fields are for. Especially, the resultId is not self-explanatory.

src/main/java/org/aksw/iguana/cc/query/handler/QueryHandler.java
```diff
@@ -141,18 +144,18 @@ public String value() {
...
+    public record QueryStreamWrapper(int index, boolean cached, Supplier<InputStream> queryInputStreamSupplier, boolean update, Integer resultId) {}
```
bigerl, line 170: same here

src/main/java/org/aksw/iguana/cc/query/handler/QueryHandler.java
```diff
@@ -161,7 +164,10 @@ public record QueryStreamWrapper(int index, boolean cached, Supplier<InputStream
     final protected Config config;
 
     final protected QueryList queryList;
-    final protected List<QueryData> queryData;
+    protected List<QueryData> queryData;
+
+    int executableQueryCount = 0;     // stores the number of queries that can be executed
+    int representativeQueryCount = 0; // stores the number of queries that are represented in the results
```
bigerl:
```suggestion
    int representedQueryCount = 0; // stores the number of queries that are represented in the results
```

src/main/java/org/aksw/iguana/cc/query/handler/QueryHandler.java
```diff
@@ -191,54 +197,139 @@ public QueryHandler(Config config) throws IOException {
...
+        AtomicInteger templateIndex = new AtomicInteger(0); // index of the next template
+        AtomicInteger index = new AtomicInteger(0);      // index of the current query
+        AtomicInteger instanceId = new AtomicInteger(0); // id of the current instance for the current template
+        queryData = templateData.queries.stream().map(
+                query -> {
+                    // If "individualResults" is turned on, move the query templates outside the range of
+                    // "representativeQueryCount" to avoid them being represented in the results.
+                    // Otherwise, if "individualResults" is turned off, the instances need to be moved outside the range
+                    // of "representativeQueryCount", but because "instantiateTemplateQueries" already appends the
+                    // instances to the end of the original queries, this will already be done.
```
bigerl: Please double check and rephrase. It sounds like it must be moved `outside` either way.

src/main/java/org/aksw/iguana/cc/query/handler/QueryHandler.java
```diff
@@ -191,54 +197,139 @@ public QueryHandler(Config config) throws IOException {
...
+
+                    // once the template instances start, the template index is reset and reused for the instances
+                    // to track to which template the instances belong
```
bigerl:
```suggestion
                    // track where the template instances belong
```

src/main/java/org/aksw/iguana/cc/query/handler/QueryHandler.java
```diff
@@ -191,54 +197,139 @@ public QueryHandler(Config config) throws IOException {
...
         }
-        return (config.caching()) ?
-                new FileCachingQueryList(querySource) : // if caching is enabled, cache the instances
-                new FileReadingQueryList(querySource);  // if caching is disabled, read the instances from the file every time
+
+        // initialize queryData based on the template data
+        AtomicInteger templateIndex = new AtomicInteger(0); // index of the next template
+        AtomicInteger index = new AtomicInteger(0);      // index of the current query
+        AtomicInteger instanceId = new AtomicInteger(0); // id of the current instance for the current template
+        queryData = templateData.queries.stream().map(
```
bigerl: I do not really get what is happening in here. That needs some more documentation. Might be that updating the comment in 282 is enough to clarify.

src/main/java/org/aksw/iguana/cc/query/handler/QueryHandler.java
```diff
@@ -191,54 +197,139 @@ public QueryHandler(Config config) throws IOException {
...
+
+                    if (index.get() >= templateData.instanceStart) {
+                        // query is an instance of a template
+
+                        // if the instance id is equal to the number of instances for the current template,
+                        // the next template is used
+                        if (instanceId.get() == templateData.instanceNumber[templateIndex.get()]) {
+                            templateIndex.getAndIncrement();
+                            instanceId.set(0);
+                        }
```
bigerl: Where is instanceId ever increased? It seems to be 0 all the time. 
Also, this looks a bit like the whole if statement might require a lock for synchronization.

src/main/java/org/aksw/iguana/cc/query/handler/QueryHandler.java
```diff
@@ -191,54 +197,139 @@ public QueryHandler(Config config) throws IOException {
...
                         writer.newLine();
                     }
                 }
-                // create a new query source based on the new instance file
-                querySource = createQuerySource(instancePath);
-            } else {
-                // query source isn't necessary, because queries aren't stored in a file,
-                // directly return a list of the instances instead
-                return new StringListQueryList(instances);
             }
```
bigerl: Please extract reading and writing template headers. Best would be a separate file. 

The same applies probably to template data in general. It currently makes the QueryHandler.java file hard to follow. Would be better to refactor and extract the handling of templates into a separate file.

src/main/java/org/aksw/iguana/cc/query/QueryData.java
```diff
@@ -12,9 +13,15 @@
...
-public record QueryData(int queryId, boolean update) {
+public record QueryData(int queryId, QueryType type, Integer templateId) {
+    public enum QueryType {
+        DEFAULT,
+        UPDATE,
+        TEMPLATE,
+        TEMPLATE_INSTANCE
+    }
+
     public static List<QueryData> generate(Collection<InputStream> queries) {
```
bigerl, line 33: add a docstring that explictly states that only default and update queries can be processed.

src/main/java/org/aksw/iguana/cc/query/QueryData.java
```diff
@@ -12,9 +13,15 @@
...
  * @param queryId The id of the query
- * @param update  If the query is an update query
  */
-public record QueryData(int queryId, boolean update) {
+public record QueryData(int queryId, QueryType type, Integer templateId) {
+    public enum QueryType {
+        DEFAULT,
+        UPDATE,
+        TEMPLATE,
+        TEMPLATE_INSTANCE
```
bigerl, line 22: Shoudn't this be 
```suggestion
        TEMPLATE_INSTANCE_DEFAULT,
        TEMPLATE_INSTANCE_UPDATE
```

src/main/java/org/aksw/iguana/cc/tasks/impl/StresstestResultProcessor.java
```diff
@@ -128,7 +128,8 @@ public void calculateAndSaveMetrics(Calendar start, Calendar end) {
             m.add(workerRes, RDF.type, IONT.worker);
             m.add(workerRes, IPROP.workerID, toInfinitePrecisionIntegerLiteral(worker.getWorkerID()));
             m.add(workerRes, IPROP.workerType, ResourceFactory.createTypedLiteral(worker.getClass().getSimpleName()));
-            m.add(workerRes, IPROP.noOfQueries, toInfinitePrecisionIntegerLiteral(config.queries().getQueryCount()));
+            // TODO: is it executable or representative?
```
bigerl: depends on `individualResults = false/true/

src/main/java/org/aksw/iguana/cc/query/QueryData.java
```diff
@@ -25,8 +32,24 @@ public static List<QueryData> generate(Collection<InputStream> queries) {
...
-            queryData.add(new QueryData(i++, update));
+            queryData.add(new QueryData(i++, update ? QueryType.UPDATE : QueryType.DEFAULT, null));
+            try {
+                query.close();
+            } catch (IOException ignored) {}
         }
         return queryData;
     }
+
+    public static boolean checkUpdate(InputStream query) {
```
bigerl: add a docstring

src/main/java/org/aksw/iguana/cc/query/handler/QueryHandler.java
```diff
@@ -161,7 +164,10 @@ public record QueryStreamWrapper(int index, boolean cached, Supplier<InputStream
     final protected Config config;
 
     final protected QueryList queryList;
-    final protected List<QueryData> queryData;
+    protected List<QueryData> queryData;
+
+    int executableQueryCount = 0;     // stores the number of queries that can be executed
```
bigerl: mention in the docstrings of executableQueryCount and representedQueryCount when they are the same (no template or setting explicitly  template's `individualResults: true`) and how these differ for templates.

src/main/java/org/aksw/iguana/cc/query/handler/QueryHandler.java
bigerl: give them a clear visibility instead of implicit package private.

bigerl (changes requested):
Thx for the PR. The changes to introduce the aggregation currently seem quite complex. See my comments for suggestions on how the code could be untangled a bit.

src/main/java/org/aksw/iguana/cc/query/QueryData.java
nck-mlcnv: No, I don't think so. `TEMPLATE` stands for the query template.

graalvm/suite.yml
bigerl: We've just had an issue with the wikidata endpoint because it requires the [User-Agent](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent) to be set, see: https://foundation.wikimedia.org/wiki/Policy:User-Agent_policy

nck-mlcnv: The Apache HttpClient automatically sends that header.

src/main/java/org/aksw/iguana/cc/query/QueryData.java
nck-mlcnv: nvm

nck-mlcnv: actually, there is nothing that really supports update queries as templates, so i don't know if that was really necessary

src/main/java/org/aksw/iguana/cc/query/handler/QueryHandler.java
nck-mlcnv: Why should it be synchronized? There is no concurrency happening.

src/main/java/org/aksw/iguana/cc/query/handler/QueryHandler.java
nck-mlcnv: I haven't extracted the header writing part to a separate file yet.

nck-mlcnv:
The version check test seems to be broken, I have to fix it soon. Also I haven't extracted the header writing part to a separate file yet.

src/main/java/org/aksw/iguana/cc/query/handler/TemplateHandler.java
```diff
@@ -0,0 +1,279 @@
...
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Random;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+import java.util.stream.Stream;
+
+public class TemplateHandler {
```
bigerl: Please add a docstring and rename to QueryTemplateManager

bigerl (changes requested):
Thx for the overhaul. Code looks good now. Not much to add. Only this small documentation and renaming request below. :)

PR #285, opened 2024-10-01
Single point version definition

nck-mlcnv:
### Versioning:
* Relocated `schema/iguana.owx` to `src/main/resources/iguana.owx` and updated the `versionIRI` to use the `${revision}` placeholder (ontology version will be equal to the version of IGUANA)
* Updated `graalvm/generate-profile.sh` to remove hardcoded version numbers and use dynamic naming for instrumented and PGO builds.
* Added `iguana.owx` to the resources section in `pom.xml` for proper inclusion during the build process.

### Deployment Workflow Updates:
* Added steps to reuse artifacts from previous jobs for the ontology

### Cleanup
* Removed old schema images


pom.xml
```diff
@@ -220,6 +220,7 @@
                                     <directory>${project.basedir}/src/main/resources/</directory>
                                     <includes>
                                         <include>start-iguana.sh</include>
+                                        <include>iguana.owx</include>
```
bigerl, line 223: As just discussed, please have a separate version for the ontology. Please Also add an action that checks if the ontology was changed. If it was changed it should require to the ontology version to be raised and to be equal to the code version.

bigerl, line 223: Please also add the ontology version to `iguana --version`. Not sure if we defined the option yet. If not, please add.

bigerl (changes requested):
Thx for the PR. Changes look good. There is only the part about independent versioning of IGUANA binaries and ontology.

src/main/java/org/aksw/iguana/cc/controller/MainController.java
```diff
@@ -74,4 +89,14 @@ public static void main(String[] argc) {
         System.exit(0);
     }
 
+    private static void outputVersion() throws IOException {
+        ClassLoader classloader = MainController.class.getClassLoader();
+        String properties = new String(Objects.requireNonNull(classloader.getResourceAsStream("version.properties")).readAllBytes());
```
bigerl, line 94: Have you checked that this also works with the native binary?

src/main/java/org/aksw/iguana/cc/controller/MainController.java
```diff
@@ -74,4 +89,14 @@ public static void main(String[] argc) {
...
+        String[] lines = properties.split("\\n");
+        String projectVersion = lines[0].split("=")[1];
+        String ontologyVersion = lines[1].split("=")[1];
+        System.out.println("IGUANA version: " + projectVersion);
```
bigerl, line 98: Please make the output match the following pattern: `IGUANA version 4.1.0 (result ontology version 4.0.0)`

bigerl (changes requested):
Thx for the PR. Changes look good to me. I just have two small comments.

bigerl (approved):
Looks good now. Ready to merge.

PR #286, opened 2024-10-01
Make linearly executing workers start at different positions

nck-mlcnv:
This pull request makes changes so that multiple workers that execute the same queries linearly, start at different positions.
The starting positions are equally distanced from each other.

### Query Handling Enhancements:

* [`src/main/java/org/aksw/iguana/cc/query/handler/QueryHandler.java`](diffhunk://#diff-9a5e1eadd71482cada7feae9ee933dd12c62112a8c5548e4c1c86c7b99176bfbR160): Added a `totalWorkerCount` field and a method to set this value. Updated the `getQuerySelectorInstance` method to use this field for the `LinearQuerySelector` initialization. [[1]](diffhunk://#diff-9a5e1eadd71482cada7feae9ee933dd12c62112a8c5548e4c1c86c7b99176bfbR160) [[2]](diffhunk://#diff-9a5e1eadd71482cada7feae9ee933dd12c62112a8c5548e4c1c86c7b99176bfbR190-R193) [[3]](diffhunk://#diff-9a5e1eadd71482cada7feae9ee933dd12c62112a8c5548e4c1c86c7b99176bfbL241-R246)

### Query Selector Improvements:

* [`src/main/java/org/aksw/iguana/cc/query/selector/impl/LinearQuerySelector.java`](diffhunk://#diff-6a76323ed269971f62258de495a664ea785020c4666cf53a4d25375aaf376048L17-R40): Modified the `LinearQuerySelector` to support an optional starting index and adjusted the `getNextIndex` and `getCurrentIndex` methods accordingly.

### Worker Initialization Fixes:

* [`src/main/java/org/aksw/iguana/cc/tasks/impl/Stresstest.java`](diffhunk://#diff-ccb14858c9c4dcce36017cf028fdf99a0a5f7e9681f1d681bed577a5247104a8R5): Updated the `Stresstest` class to initialize query handlers and count the number of workers for each query handler during both warmup and main execution phases. [[1]](diffhunk://#diff-ccb14858c9c4dcce36017cf028fdf99a0a5f7e9681f1d681bed577a5247104a8R5) [[2]](diffhunk://#diff-ccb14858c9c4dcce36017cf028fdf99a0a5f7e9681f1d681bed577a5247104a8R48-R56) [[3]](diffhunk://#diff-ccb14858c9c4dcce36017cf028fdf99a0a5f7e9681f1d681bed577a5247104a8R67-R75) [[4]](diffhunk://#diff-ccb14858c9c4dcce36017cf028fdf99a0a5f7e9681f1d681bed577a5247104a8L86-L89)

### Bug Fixes:

* [`src/main/java/org/aksw/iguana/cc/worker/impl/SPARQLProtocolWorker.java`](diffhunk://#diff-de29e8baf834d272b515e906813d739f2e310324169311d147df54ca5897786dL273-R273): Fixed a bug in the `executeHttpRequest` method to correctly use the `querySelector`'s current index.

### Testing Enhancements:

* [`src/test/java/org/aksw/iguana/cc/query/selector/impl/LinearQuerySelectorTest.java`](diffhunk://#diff-e5138a189b2a80b197a1a2618a5b286b370a56abdb26bfe5653eb7e3427e65daR29-R42): Added a test case to verify the functionality of the `LinearQuerySelector` with a starting index.

### DIFF
--- a/src/main/java/org/aksw/iguana/cc/query/handler/QueryHandler.java
+++ b/src/main/java/org/aksw/iguana/cc/query/handler/QueryHandler.java
@@ -157,6 +157,7 @@ public class QueryHandler {
     final protected QueryList queryList;
 
     private int workerCount = 0; // give every worker inside the same worker config an offset seed
+    private int totalWorkerCount = 0;
 
     final protected int hashCode;
 
@@ -184,6 +185,10 @@ public class QueryHandler {
                     new FileReadingQueryList(querySource);
         }
         this.hashCode = queryList.hashCode();
+    }
+
+    public void setTotalWorkerCount(int workers) {
+        this.totalWorkerCount = workers;
     }
 
     private QueryList initializeTemplateQueryHandler(QuerySource templateSource) throws IOException {
@@ -238,7 +243,7 @@ public class QueryHandler {
 
     public QuerySelector getQuerySelectorInstance() {
         switch (config.order()) {
-            case LINEAR -> { return new LinearQuerySelector(queryList.size()); }
+            case LINEAR -> { return new LinearQuerySelector(queryList.size(), totalWorkerCount != 0 ? (queryList.size() * workerCount++) / totalWorkerCount : 0); }
             case RANDOM -> { return new RandomQuerySelector(queryList.size(), config.seed() + workerCount++); }
         }
 
--- a/src/main/java/org/aksw/iguana/cc/query/selector/impl/LinearQuerySelector.java
+++ b/src/main/java/org/aksw/iguana/cc/query/selector/impl/LinearQuerySelector.java
@@ -32,6 +32,7 @@ public class LinearQuerySelector extends QuerySelector {
 
     /**
      * Return the current index. This is the index of the last returned query.
+     * If no query was returned yet, the method will return -1.
      *
      * @return the current index
      */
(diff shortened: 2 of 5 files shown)


src/main/java/org/aksw/iguana/cc/query/handler/QueryHandler.java
```diff
@@ -238,7 +243,7 @@ private QuerySource createQuerySource(Path path) throws IOException {
 
     public QuerySelector getQuerySelectorInstance() {
         switch (config.order()) {
-            case LINEAR -> { return new LinearQuerySelector(queryList.size()); }
+            case LINEAR -> { return new LinearQuerySelector(queryList.size(), totalWorkerCount != 0 ? (queryList.size() * workerCount++) / totalWorkerCount : 0); }
```
bigerl, line 246: When would `totalWorkerCount != 0` be true? 
Should this not be `totalWorkerCount != 1`?

bigerl (changes requested):
Thx for the PR. Changes look mostly good. There is just one spot where I am a little confused. Only, I am not too fond of the  -1 initialization of `getCurrentIndex()`. Still, I don't have a better solution at hand ATM.

src/main/java/org/aksw/iguana/cc/query/handler/QueryHandler.java
nck-mlcnv: `totalWorkerCount` is zero during testing and you would get a lot of division by zero errors, so I added that condition to the method. It would also make that method more robust, so I just added that condition instead of changing every failing test.

bigerl (approved):

